PR #12, opened 2023-07-08
element (v5)

rvilarl:
Here we go...

`Element` is the base of vexflow 5. An `Element` has a member `text` which is a series of unicode characters (including SmufL codes). This text is drawn using the `textFont` which is normally a comma separated list of fonts. Element calls internally `measureText` to calculate the `metrics`, `height` and `ẃidth` of the `text`

`renderText` will render the `text` using the context provided at the coordinates provided. Internally `Element` will apply the `xShift` and `yShift`. 

So far no visual differences. The default fonts are moved to `commonMetrics`.

### DIFF
--- a/src/element.ts
+++ b/src/element.ts
@@ -3,11 +3,12 @@
 // @author Mohit Cheppudira
 
 import { BoundingBox } from './boundingbox';
-import { Font, FontInfo, FontStyle, FontWeight } from './font';
+import { Font, FontInfo } from './font';
 import { Registry } from './registry';
 import { RenderContext } from './rendercontext';
+import { Tables } from './tables';
 import { Category } from './typeguard';
-import { defined, prefix } from './util';
+import { defined, prefix, RuntimeError } from './util';
 
 /** Element attributes. */
 export interface ElementAttributes {
@@ -59,12 +60,11 @@ export interface ElementStyle {
  * Element implements a generic base class for VexFlow, with implementations
  * of general functions and properties that can be inherited by all VexFlow elements.
  *
- * The Element is an abstract class that needs to be subclassed to work. It handles
- * style and text-font properties for the Element and any child elements, along with
- * working with the Registry to create unique ids, but does not have any tools for
- * formatting x or y positions or connections to a Stave.
+ * The Element handles style and text-font properties for the Element and any child
+ * elements, along with working with the Registry to create unique ids, but does not
+ * have any tools for formatting x or y positions or connections to a Stave.
  */
-export abstract class Element {
+export class Element {
   static get CATEGORY(): string {
     return Category.Element;
   }
@@ -77,17 +77,6 @@ export abstract class Element {
     return `auto${Element.ID++}`;
   }
 
-  /**
-   * Default font for text. This is not related to music engraving. Instead, see `Flow.setMusicFont(...fontNames)`
-   * to customize the font for musical symbols placed on the score.
-   */
-  static TEXT_FONT: Required<FontInfo> = {
-    family: Font.SANS_SERIF,
-    size: Font.SIZE,
-    weight: FontWeight.NORMAL,
-    style: FontStyle.NORMAL,
-  };
-
   #context?: RenderContext;
   #attrs: ElementAttributes;
 
@@ -97,20 +86,38 @@ export abstract class Element {
   protected registry?: Registry;
 
   /**
-   * Some elements include text.
    * The `textFont` property contains information required to style the text (i.e., font family, size, weight, and style).
-   * It is undefined by default, and can be set using `setFont(...)` or `resetFont()`.
    */
-  protected textFont?: Required<FontInfo>;
+  protected textFont: Required<FontInfo>;
+  protected text = '';
+  protected textMetrics: TextMetrics = {
+    fontBoundingBoxAscent: 0,
+    fontBoundingBoxDescent: 0,
+    actualBoundingBoxAscent: 0,
+    actualBoundingBoxDescent: 0,
+    actualBoundingBoxLeft: 0,
+    actualBoundingBoxRight: 0,
+    width: 0,
+  };
 
-  constructor() {
+  protected width: number = 0;
+  protected yShift: number = 0;
+  protected xShift: number = 0;
+
+  constructor(category?: string) {
     this.#attrs = {
       id: Element.newID(),
-      type: this.getCategory(),
+      type: category ? category : (<typeof Element>this.constructor).CATEGORY,
       class: '',
     };
 
     this.rendered = false;
+    this.textFont = {
+      family: Tables.lookupMetric(`${this.#attrs.type}.fontFamily`),
+      size: Tables.lookupMetric(`${this.#attrs.type}.fontSize`),
+      weight: Tables.lookupMetric(`${this.#attrs.type}.fontWeight`),
+      style: Tables.lookupMetric(`${this.#attrs.type}.fontStyle`),
+    };
 
     // If a default registry exist, then register with it right away.
     Registry.getDefaultRegistry()?.register(this);
@@ -131,7 +138,7 @@ export abstract class Element {
   }
 
   getCategory(): string {
-    return (<typeof Element>this.constructor).CATEGORY;
+    return `${this.#attrs.type}`;
   }
 
   /**
@@ -213,8 +220,9 @@ export abstract class Element {
   }
 
   /** Draw an element. */
-  // eslint-disable-next-line
-  abstract draw(...args: any[]): void;
+  draw(): void {
+    throw new RuntimeError('Element', 'Draw not defined');
+  }
 
   /** Check if it has a class label (An element can have multiple class labels).  */
   hasClass(className: string): boolean {
@@ -351,12 +359,15 @@ export abstract class Element {
    * Each Element subclass may specify its own default by overriding the static `TEXT_FONT` property.
    */
   setFont(font?: string | FontInfo, size?: string | number, weight?: string | number, style?: string): this {
-    // Allow subclasses to override `TEXT_FONT`.
-    const defaultTextFont: Required<FontInfo> = (<typeof Element>this.constructor).TEXT_FONT;
+    const defaultTextFont: Required<FontInfo> = {
+      family: Tables.lookupMetric(`${this.#attrs.type}.fontFamily`),
+      size: Tables.lookupMetric(`${this.#attrs.type}.fontSize`),
+      weight: Tables.lookupMetric(`${this.#attrs.type}.fontWeight`),
+      style: Tables.lookupMetric(`${this.#attrs.type}.fontStyle`),
+    };
 
     const fontIsObject = typeof font === 'object';
     const fontIsString = typeof font === 'string';
-    const fontIsUndefined = font === undefined;
     const sizeWeightStyleAreUndefined = size === undefined && weight === undefined && style === undefined;
 
     if (fontIsObject) {
@@ -365,13 +376,7 @@ export abstract class Element {
     } else if (fontIsString && sizeWeightStyleAreUndefined) {
       // `font` is case 2) CSS font shorthand.
       this.textFont = Font.fromCSSString(font);
-    } else if (fontIsUndefined && sizeWeightStyleAreUndefined) {
-      // All arguments are undefined. Do not check for `arguments.length === 0`,
-      // which fails on the edge case: `setFont(undefined)`.
-      // TODO: See if we can remove this case entirely without introducing a visual diff.
-      // The else case below seems like it should be equivalent to this case.
-      this.textFont = { ...defaultTextFont };
-    } else {
+     } else {
       // `font` is case 3) a font family string (e.g., 'Times New Roman').
       // The other parameters represent the size, weight, and style.
       // It is okay for `font` to be undefined while one or more of the other arguments is provided.
@@ -383,6 +388,7 @@ export abstract class Element {
         style ?? defaultTextFont.style
       );
     }
+    this.measureText();
     return this;
   }
 
@@ -391,28 +397,14 @@ export abstract class Element {
    * 'bold 10pt Arial'.
    */
   getFont(): string {
-    if (!this.textFont) {
-      this.resetFont();
-    }
     return Font.toCSSString(this.textFont);
-  }
-
-  /**
-   * Reset the text font to the style indicated by the static `TEXT_FONT` property.
-   * Subclasses can call this to initialize `textFont` for the first time.
-   */
-  resetFont(): void {
-    this.setFont();
   }
 
   /** Return a copy of the current FontInfo object. */
   get fontInfo(): Required<FontInfo> {
-    if (!this.textFont) {
-      this.resetFont();
-    }
     // We can cast to Required<FontInfo> here, because
     // we just called resetFont() above to ensure this.textFont is set.
-    return { ...this.textFont } as Required<FontInfo>;
+    return this.textFont;
   }
 
   set fontInfo(fontInfo: FontInfo) {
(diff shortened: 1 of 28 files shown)


ronyeh:
Hi Rodrigo!

I just started the review, and I expect it'll take a few days. While you have been thinking about this problem for a long time, the rest of us haven't, so expect some dumb questions...

Also, the more people you can convince to provide input on your PRs the better, I think. It is OK if the end product doesn't satisfy everyone's requests, but just knowing what others think will help guide your design decision better.

Ron

src/element.ts
```diff
@@ -77,17 +77,6 @@ export abstract class Element {
...
-   * Default font for text. This is not related to music engraving. Instead, see `Flow.setMusicFont(...fontNames)`
-   * to customize the font for musical symbols placed on the score.
-   */
-  static TEXT_FONT: Required<FontInfo> = {
-    family: Font.SANS_SERIF,
-    size: Font.SIZE,
-    weight: FontWeight.NORMAL,
-    style: FontStyle.NORMAL,
-  };
-
```
ronyeh, line 90: In VexFlow 4.x, we had fonts for styling "text-like" elements, and fonts for styling "music symbols". `TEXT_FONT` was used for the "text-like" elements.... i.e., anything that could accept CSS style font declarations (e.g., '15px Arial, sans-serif;`).

By deleting TEXT_FONT, are you saying that going forward, all elements, whether they are "text-like" or "music symbols" will all go through the same path for font styling? They will all use CSS-style font declarations?

src/element.ts
```diff
@@ -97,20 +86,38 @@ export abstract class Element {
...
+  protected width: number = 0;
+  protected yShift: number = 0;
+  protected xShift: number = 0;
 
-  constructor() {
+  constructor(category?: string) {
     this.#attrs = {
       id: Element.newID(),
-      type: this.getCategory(),
+      type: category ? category : (<typeof Element>this.constructor).CATEGORY,
```
ronyeh, line 110: Can this line use the nullish coalescing operator?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

```
type: category ?? (<typeof Element>this.constructor).CATEGORY,
```

src/element.ts
```diff
@@ -213,8 +220,9 @@ export abstract class Element {
   }
 
   /** Draw an element. */
-  // eslint-disable-next-line
-  abstract draw(...args: any[]): void;
+  draw(): void {
+    throw new RuntimeError('Element', 'Draw not defined');
```
ronyeh, line 224: I notice Element is no longer an abstract class. That's fine.... but here the draw() method just throws a RuntimeError. So.... Element is still basically an abstract class that requires draw() to be overridden? 

Or should `draw()` have a basic implementation that just draws `this.text`?

src/tables.ts
```diff
@@ -586,6 +718,41 @@ export class Tables {
...
   }
 
+  /**
+   * Use the provided key to look up a value in this font's metrics file (e.g., bravuraMetrics.ts, petalumaMetrics.ts).
+   * @param key is a string separated by periods (e.g., stave.endPaddingMax, clef.lineCount.'5'.shiftY).
+   * @param defaultValue is returned if the lookup fails.
+   * @returns the retrieved value (or `defaultValue` if the lookup fails).
+   */
+  // eslint-disable-next-line
+  static lookupMetric(key: string, defaultValue?: Record<string, any> | number): any {
```
ronyeh, line 728: Was this previously in the Font class as `Font.lookupMetric()`?

I'm a bit worried about tables.ts becoming a huge catch-all class that handles everything.

Should the `CommonMetrics` class be in its own file? It could be called the `Metrics` class, and placed in a `metrics.ts` file. All the CommonMetrics serve as the defaults that can be overridden by other fonts.

Then this lookupMetric(...) function can be inside metrics.ts as Metrics.lookup().

Anyways, just a thought. If you believe that the Tables is in fact the Metrics class, then it's OK to keep it all here. I just want to make sure one class isn't handling everything.

src/tables.ts
```diff
@@ -9,6 +9,138 @@ import { RuntimeError } from './util';
...
+  },
+
+  Strokes: {
+    text: {
+      fontFamily: 'Times New Roman, serif',
+      fontSize: 10,
+      fontStyle: 'italic',
+      fontWeight: 'bold',
+    },
+  },
```
ronyeh, line 108: It's interesting that all other metrics have fontFamily, fontSize, fontStyle, fontWeight defined on the Element type. 

Why is `Strokes.text` special? Why can't those properties be defined under Strokes directly?

src/strokes.ts
```diff
@@ -243,8 +233,14 @@ export class Stroke extends Modifier {
 
     // Draw the rasquedo "R"
     if (this.type === Stroke.Type.RASQUEDO_DOWN || this.type === Stroke.Type.RASQUEDO_UP) {
+      const textFont = {
+        family: Tables.lookupMetric(`Strokes.text.fontFamily`),
+        size: Tables.lookupMetric(`Strokes.text.fontSize`),
+        weight: Tables.lookupMetric(`Strokes.text.fontWeight`),
+        style: Tables.lookupMetric(`Strokes.text.fontStyle`),
+      };
```
ronyeh, line 241: Is this because Strokes will have a different fontFamily than Strokes.text?

For example, Strokes will use `Bravura` while Strokes.text will use `12pt Arial`?

src/tables.ts
```diff
@@ -9,6 +9,138 @@ import { RuntimeError } from './util';
 
 const RESOLUTION = 16384;
 
+export const CommonMetrics = {
+  fontFamily: 'Bravura',
+  fontSize: 30,
+  fontWeight: 'normal',
+  fontStyle: 'normal',
+
```
ronyeh, line 17: This block at the top. Does it serve as default values for fontFamily, fontSize, etc?

I find it interesting that we are now combining text font and music font handling into one. That's cool... but it does seem odd that FretHandFinger's `Arial, sans-serif` will override the default `Bravura`.... since they very different. Arial is for rendering text, while Bravura is for rendering music glyphs.

tests/stave_tests.ts
```diff
@@ -747,10 +747,10 @@ function configureAllLines(options: TestOptions, contextBuilder: ContextBuilder)
...
   const ctx = contextBuilder(options.elementId, 900, 140);
   const stave = new Stave(300, 10, 300);
-  stave.setText('Violin', Modifier.Position.LEFT);
-  stave.setText('Right Text', Modifier.Position.RIGHT);
-  stave.setText('Above Text', Modifier.Position.ABOVE);
-  stave.setText('Below Text', Modifier.Position.BELOW);
+  stave.setStaveText('Violin', Modifier.Position.LEFT);
+  stave.setStaveText('Right Text', Modifier.Position.RIGHT);
+  stave.setStaveText('Above Text', Modifier.Position.ABOVE);
+  stave.setStaveText('Below Text', Modifier.Position.BELOW);
```
ronyeh, line 753: Can you explain why it is now renamed to setStaveText?

It seems a bit redundant to say `stave.setStaveText(...)`. Why not just `stave.setText(...)`?

tests/font_tests.ts
```diff
@@ -46,29 +46,24 @@ function setFont(assert: Assert): void {
...
-  assert.equal(flat.getFont(), 'bold 10pt Arial, sans-serif');
-  flat.setFont(undefined, undefined, FontWeight.BOLD, FontStyle.ITALIC);
-  assert.equal(flat.getFont(), 'italic bold 10pt Arial, sans-serif');
-  flat.setFont(undefined, undefined, FontWeight.BOLD, 'oblique');
-  assert.equal(flat.getFont(), 'oblique bold 10pt Arial, sans-serif');
+  flat.setFont(undefined, undefined, 'bold', undefined);
+  assert.equal(flat.getFont(), 'bold 30pt Bravura,Gonville,Custom');
+  flat.setFont(undefined, undefined, 'bold', 'italic');
+  assert.equal(flat.getFont(), 'italic bold 30pt Bravura,Gonville,Custom');
+  flat.setFont(undefined, undefined, 'bold', 'oblique');
```
ronyeh, line 62: Nice. So we now can specify music fonts as CSS font strings.

What about the text-like elements. Do we have any tests that verify `'10pt Arial, sans-serif'` still works and parses correctly? I don't see why it wouldn't, since it's all strings, but we might as well keep the test for Arial around, in case any of our elements uses Arial as their main text font.

src/stave.ts
```diff
@@ -328,7 +319,7 @@ export class Stave extends Element {
   }
 
   // Text functions
-  setText(
+  setStaveText(
```
ronyeh, line 322: Was this renamed because it conflicts with the new Element.text() method? Makes sense.

src/element.ts
```diff
@@ -97,20 +86,38 @@ export abstract class Element {
   protected registry?: Registry;
 
   /**
-   * Some elements include text.
    * The `textFont` property contains information required to style the text (i.e., font family, size, weight, and style).
-   * It is undefined by default, and can be set using `setFont(...)` or `resetFont()`.
    */
-  protected textFont?: Required<FontInfo>;
+  protected textFont: Required<FontInfo>;
+  protected text = '';
```
ronyeh, line 92: Element.text seems like the new important addition. Can we add a good comment for this property? Some of your description of this PR would be useful here. Basically, every Element that is placed on the staff is fundamentally a `text: string` styled by a SMuFL font.

src/element.ts
```diff
@@ -131,7 +138,7 @@ export abstract class Element {
   }
 
   getCategory(): string {
-    return (<typeof Element>this.constructor).CATEGORY;
+    return `${this.#attrs.type}`;
```
ronyeh, line 141: Can we just return `this.#attrs.type` here? Isn't it already a string?

src/element.ts
```diff
@@ -491,4 +483,95 @@ export abstract class Element {
...
+  }
+
+  /** Set element width. */
+  setWidth(width: number): this {
+    this.width = width;
+    return this;
+  }
+
+  /** Shift element down `y` pixels. Negative values shift up. */
+  setYShift(y: number): this {
```
ronyeh, line 499: Should the parameter name be `dY` or `deltaY`? (or I suppose it could also be `yShift`)?

src/element.ts
```diff
@@ -491,4 +483,95 @@ export abstract class Element {
...
+
+  /** Get shift element `y` */
+  getYShift(): number {
+    return this.yShift;
+  }
+
+  /**
+   * Set shift element right `x` pixels. Negative values shift left.
+   */
+  setXShift(x: number): this {
```
ronyeh, line 512: Same as with yShift, I think maybe it should be `dX` or `xShift`.

src/element.ts
```diff
@@ -491,4 +483,95 @@ export abstract class Element {
...
+
+  /** Get shift element `x` */
+  getXShift(): number {
+    return this.xShift;
+  }
+
+  /**
+   * Set element text
+   */
+  setText(txt: string): this {
```
ronyeh, line 525: Why not `text`? Or if you are trying to be concise, you coulde use `t` :-)

src/element.ts
```diff
@@ -491,4 +483,95 @@ export abstract class Element {
...
+   */
+  renderText(ctx: RenderContext, xPos: number, yPos: number): void {
+    ctx.save();
+    ctx.setFont(this.textFont);
+    ctx.fillText(this.text, xPos + this.xShift, yPos + this.yShift);
+    ctx.restore();
+  }
+
+  /** Canvas used to measure text */
+  protected static txtCanvas?: HTMLCanvasElement;
```
ronyeh, line 546: Is this an internal member? Should it use the private `#` syntax? See fraction.ts:

```
static #fractionA = new Fraction();
```

AaronDavidNewman:
@rvilarl is there a branch that has all the vexflow music font changes on it?  These are mostly refactoring changes, I assume this is in service of using the fonts to render the music instead of extracted path.  It's kind of hard to evaluate them out of context, they seem fine but the 'why' isn't clear to me.   Even if it's a private branch and not everything is ready to merge, it would be useful to me in doing the reviews.

src/element.ts
rvilarl: We do not have a x, y position to draw, this is why I left it without a basic call to renderText

src/tables.ts
rvilarl: Yes, you will see that the combination will spread over the future classes. I have overriden the defaults to make sure that there are no visual differences. After this PR, the next, chordsymbols, will come with visual differences.

src/strokes.ts
rvilarl: yes :)

src/stave.ts
rvilarl: Yes, this is right.

tests/stave_tests.ts
rvilarl: Already answered :)

tests/font_tests.ts
rvilarl: There is no check on the font famly, any combination is fine. You will see that in the HACK in flow.ts in the next PR.

src/element.ts
rvilarl: done :)

src/element.ts
rvilarl: done :)

src/element.ts
rvilarl: done

src/element.ts
rvilarl: done

src/element.ts
rvilarl: done

src/element.ts
rvilarl: Done

rvilarl:
> @rvilarl is there a branch that has all the vexflow music font changes on it? These are mostly refactoring changes, I assume this is in service of using the fonts to render the music instead of extracted path. It's kind of hard to evaluate them out of context, they seem fine but the 'why' isn't clear to me. Even if it's a private branch and not everything is ready to merge, it would be useful to me in doing the reviews.

The branch that we will review step by step is https://github.com/rvilarl/vexflow/tree/vf5/vexflow5
@AaronDavidNewman the next PR is ChordSymbols, there the concept will be clearer and we definetively need your review there.

rvilarl:
@ronyeh I will merge now, if there is something pending we can add it on the next PR. The next PR gives more insight into the vexflow5 concept.

ronyeh:
Hi Rodrigo!

I wasn't finished reviewing this PR. I needed more time to review each file.

Sorry, I should have been more clear. In the future, could you wait for a "Looks Good To Me / LGTM" before merging? Or you could provide your feedback or address the comments I made and then I can go ahead and merge.

I understand that you have been working on your approach for a while now, but to the rest of us who are also interested in VexFlow, some of the concepts and the specific approach you took are completely new information. I personally trust that you made good design decisions, but we all need just a bit more time to process this new information (and ideally, to read through each line of code you wrote).

You're a maintainer of VexFlow 5, so of course you can go ahead and merge any and all PRs you wish. But if you don't give folks like Michael, Mohit, myself, Aaron, Simon, Greg, etc a chance to comment on or review your work, VexFlow 5 might ultimately end up being "Rodrigo's branch" and everyone else might hesitate to upgrade from VexFlow 4.x.

src/element.ts
ronyeh: Thanks. This is helpful.

src/element.ts
ronyeh: So if Element cannot be used by itself, should it be an abstract class (like it was in VexFlow 4)?

If you don't declare it as `abstract`, then a developer should be able to say new Element(...) and then start using it to draw elements on the staff, correct?

src/tables.ts
ronyeh: OK, thanks. My main concern is having tables.ts grow to handle all the things.... :-)

And maybe tables.ts needs a better name? Or is "Tables" the best description for its purpose that we can come up with, hehe....

src/stave.ts
ronyeh: Can things like this be documented in the `docs/changelog/5.0.md` file? 

This breaking change might trip up someone trying to upgrade from VexFlow 4.x to 5.0. If they have code that calls setText(...) for some reason, it will now call the incorrect setText(). Instead, their code should call `setStaveText(...)`.

rvilarl:
This is easy to resolve, I will roll it back.

PR #15, opened 2023-07-10
chordsymbol (v5)

rvilarl:
This is the first complete example of vexflow 5. Each `ChordSymbolBlock` is now an `Element` and we use the `Element` infrastructure to draw them: all at the same location with their associated `xShift` or `yShift`. I believe that this has reduced the complexity of the code a lot (almost half of lines) and we can now use other fonts where we do not have the associated metrics.


rvilarl:
@ronyeh @AaronDavidNewman looking forward to get your feedback/reviews. 

There are visual differences.

rvilarl:
I have set this PR to draft to wait for the merge of #18. I will rebase then.

ronyeh:
Can you summarize what the visual differences are and why they are introduced with this PR?

Giving us the context / summary of the big idea would help. And if you can direct us to the correct files / methods and line numbers to start our review, that would help immensely!

AaronDavidNewman:
The symbols only appear in Petaluma font. Am I doing something wrong?

image

image

rvilarl:
> The symbols only appear in Petaluma font. Am I doing something wrong?

@AaronDavidNewman this is the boxy problem that @mscuthbert refered to. I forgot to update the the load of web fonts, I have pushed it. Flow.html should work now fine but to get the testing infrastructure running you might need to install the fonts in your system. Please check how it goes.

rvilarl:
Rebased and ready to review

rvilarl:
> Can you summarize what the visual differences are and why they are introduced with this PR?

The source of the visual differences are two:
- The browser is positioning now main of the characters and glyphs, they are no longer calculated by us at run time.
- ChordSymbol is now using the appropriate Text Font for each family ie.: BravuraText for Bravura,... see HACK in flow.ts
 
> Giving us the context / summary of the big idea would help. And if you can direct us to the correct files / methods and line numbers to start our review, that would help immensely!

The big idea:
Each `ChordSymbolBlock` is now an `Element` and we use the `Element` infrastructure to draw them: all at the same location with their associated `xShift` or `yShift`. I believe that this has reduced the complexity of the code a lot (almost half of lines) and we can now use other fonts where we do not have the associated metrics.

This means:
- `SymbolTypes` is no longer required, all `ChordSymbolBlock` are `text`
- `IsSuperscript`, `IsSuperscript` are now members of `ChordSymbolBlock`
-  Font specific `ChordSymbolMetrics` and `TextFormatter` are no longer required. `ChordSymbolBlock` as an extension of `Element` knows how to calculate `metrics` and `width`. `yShift` and `xShift` is also provided by `Element`
- Kerning calculations no longer required.
- `Draw` is much simpler now calling `symbol.renderText`

Hope this helps!
-

ronyeh:
Thanks!

For a future reader dropping by:
- ChordSymbols are the elements that look like `C#7` that float above the stave, indicating which chord to play.
- It looks like this: https://unpkg.com/vexflow@4.2.2/tests/flow.html#filter=ChordSymbol


@rvilarl What do you use **Bravura Text** for? Is it just for the music symbols (like sharp ♯ and flat ♭) or are you using it to style the alpha numeric characters as well?

**Bravura Text** is not for styling alpha numeric characters. Bravura Text and regular Bravura are both used to render music symbols only.

See: https://github.com/steinbergmedia/bravura/blob/master/redist/bravura-text.md

Maybe the glyph registration points are different than regular Bravura, and maybe each glyph's **advance width** is customized so that the font is easier to use in a word processor.

Bravura did not come with a special font for displaying regular text alongside Bravura scores. Instead, the designers recommend some text fonts in their [metadata.json](https://github.com/steinbergmedia/bravura/blob/02e8ed29a29115df35007d1178cebaeee26c20e1/redist/bravura_metadata.json#L28-L33) file:

https://github.com/steinbergmedia/bravura/blob/02e8ed29a29115df35007d1178cebaeee26c20e1/redist/bravura_metadata.json#L28-L33

```json
"textFontFamily":[
  "Academico",
  "Century Schoolbook",
  "Edwin",
  "serif"
],
```


Compare this with Petaluma: https://github.com/steinbergmedia/petaluma/tree/master/redist/otf

Petaluma also has a PetalumaText. Both are for displaying music symbols.

However, since Petaluma is supposed to look like handwriting, the designers also included **PetalumaScript** for displaying alphanumeric characters alongside scores engraved with Petaluma.


When @AaronDavidNewman contributed the ChordSymbol work in 2020, he and Mohit had a discussion about possible fonts to pair with the score:

https://github.com/0xfe/vexflow/pull/790#issuecomment-634589783
https://github.com/0xfe/vexflow/pull/790#issuecomment-637512786

I think they eventually arrived on Roboto Slab.

In the end, it's personal preference. But now that @sschmidTU has alerted us to the possible GDPR compliance problems with using Google Web Fonts (or any non-self-hosted font), perhaps it is easier if we just paired Bravura with with a web safe font that is included with most operating systems? 

We could use the font stack:
```
Georgia, Times, "Times New Roman", serif;
```

https://www.cssfontstack.com/Georgia
https://www.cssfontstack.com/Times-New-Roman

Then we can teach the developer how they can customize the font if they wish, using our vexflow-fonts or using google fonts, or by self hosting fonts.

What do you all think? Should we remove Roboto Slab from VexFlow by default, as long as we make it easy for the developer to self host or integrate their favorite web fonts?

ronyeh:
Here is a discussion about mixing regular text alongside Bravura symbols, with comments from the original designer:

https://forums.steinberg.net/t/ot-how-to-use-bravura-text/160206/6

ronyeh:
Here the Bravura designer mentions that Academico is also available under SIL open font license:

https://forums.steinberg.net/t/academico-open-font-license/114847

So maybe we should just extract the Academico font from Dorico and include it in vexflow-fonts.

ronyeh:
Academico: https://befonts.com/academico-serif-font.html

rvilarl:
Hi @ronyeh Thanks a lot for looking into that! The current HACK in flow takes the following combinations:
```js
      case 'Bravura':
        CommonMetrics.fontFamily = 'Bravura,BravuraText';
        break;
      case 'Gonville':
        CommonMetrics.fontFamily = 'GonvilleSmufl,Bravura,Roboto Slab';
        break;
      case 'Leland':
        CommonMetrics.fontFamily = 'Leland,Bravura,LelandText';
        break;
      case 'Petaluma':
        CommonMetrics.fontFamily = 'Petaluma,Bravura,PetalumaScript';
```
It would be great to include Academico in vexflow-fonts! What should we use with Leland. I will also make a PR now with an improved inspector of the glyphs.

ronyeh:
image

Here I demonstrate that Bravura Text doesn't include any glyphs for ASCII characters. The font family I used was:

```
font-family: 'Bravura Text', Impact;
```

And the HTML characters are:

```
<div> - &#xE050; - C&#x266D; - G&#x266E; - Em&#x266F;7</div>
```

The first character is just copy and pasted directly from the SMuFL specification webpage, so it'll probably show a box. The other music symbols use the unicode code points.

ronyeh:
image

```
font-family: 'Bravura Text', 'Academico';
```

rvilarl:
But I would rather use the combination

```css
font-family: 'Bravura, Academico';
```

`Bravura Text` is bigger, and looking into the details, it contains the same glyphs repeated at different yOffsets.

rvilarl:
I have read that the only problem with the google fonts is to download them from google because google stores the IP address of the client using the font. If we download it from vexflow-fonts it should not be a problem, I believe.

ronyeh:
image

Here is the difference when Bravura Text or Bravura are used alongside ASCII characters typeset in Academico.

Bravura Text is aligned properly when placed next to regular text. 

Plain old Bravura glyphs might take up more vertical space, and have different glyph origins.

rvilarl:
@ronyeh you are not using the right glyphs for the accidentals in the chordsymbols, you need these

https://w3c.github.io/smufl/latest/tables/standard-accidentals-for-chord-symbols.html?highlight=csym#standard-accidentals-for-chord-symbols-ued60ued6f

They are already used in this PR

ronyeh:
> I have read that the only problem with the google fonts is to download them from google because google stores the IP address of the client using the font. If we download it from vexflow-fonts it should not be a problem, I believe.

I do not know what unpkg.com does. That is not a server we control. Like jsdelivr, unpkg just makes NPM packages available over the web. We uploaded some SMuFL fonts to it so that we can load web fonts with VexFlow.

To be 100% GDPR safe, the developer would most likely need to self-host their fonts.

I have started experimenting with this. At the moment, you can npm install individual woff2 fonts as follows:

```
npm install @vexflow-fonts/bravura
npm install @vexflow-fonts/bravura-text
```

If this approach works for people, we can also upload Academico or other SIL open fonts to NPM, so that they can be installed individually. I borrowed this approach from [fontsource.org](https://fontsource.org/).

ronyeh:
> @ronyeh you are not using the right glyphs for the accidentals in the chordsymbols, you need these
> 
> https://w3c.github.io/smufl/latest/tables/standard-accidentals-for-chord-symbols.html?highlight=csym#standard-accidentals-for-chord-symbols-ued60ued6f
> 
> The hieght of the box is resolved using actualBoundingBoxAscent/Descent metrics rather than fontBoundingBoxAscent/Descent.
> 
> They are already used in this PR

OK, I was just using the glyph code points recommended by the original font designer:

https://forums.steinberg.net/t/ot-how-to-use-bravura-text/160206/6

image


In this example, I was typesetting the sharp and flat symbols next to regular Academico text. It is possible that we want to use different glyph code points when placing sharp and flat symbols next to other music symbols (note heads).

ronyeh:
I read your link. Thanks! I do see that there are special sharp and flat glyphs for use alongside chord symbols. 

I think the tricky part is what happens when chord symbols and sharp/flat symbols are used next to Academico text?

AaronDavidNewman:
5 more responses came in while I was typing.  So I'll keep it short.

glyphs need to be scaled up, they are too small.  So you will need to add the distinction back between 'text' and 'glyph', even if they come from the same font file since they will need to be in separate elements.

unpkg method won't work with local development.  You need to include the static fonts you want it to use, or put the woff in a cdn and reference them statically in the html file.  Otherwise you will be fielding questions forever about why xxx doesn't show up in my computer.

Other than the issue with loading fonts and the scale, I think this scheme will work for chord symbols.

ronyeh:
> unpkg method won't work with local development. You need to include the static fonts you want it to use, or put the woff in a cdn and reference them statically in the html file. Otherwise you will be fielding questions forever about why xxx doesn't show up in my computer.

Thanks @AaronDavidNewman 

Can you clarify a abit? When you say put the woff in a CDN.... that's what we are trying to do with the unpkg.com (which is a CDN for serving NPM packages).

For example, here is the Bravura woff2: https://unpkg.com/vexflow-fonts@1.0.5/bravura/Bravura_1.392.woff2

You can use the **@font-face** CSS declaration and point to that URL.

VexFlow also provides a JS way to do the same thing. The FontFace JS API is equivalent to the CSS `@font-face` method.

`Font.loadWebFont(fontName: string, woffURL: string, includeWoff2?: boolean)`

```
  static async loadWebFont(fontName: string, woffURL: string, includeWoff2: boolean = true): Promise<FontFace> {
    const woff2URL = includeWoff2 ? `url(${woffURL}2) format('woff2'), ` : '';
    const woff1URL = `url(${woffURL}) format('woff')`;
    const woffURLs = woff2URL + woff1URL;
    const fontFace = new FontFace(fontName, woffURLs);
    await fontFace.load();
    document.fonts.add(fontFace);
    return fontFace;
  }
```

Regarding glyph vs text sizes, I think that's why the sizing and registration points are different in Bravura Text. When symbols are placed directly next to ASCII text, they need to be aligned and sized correctly.

But if we want to just use Bravura (my preference, so we don't have to load multiple fonts), then we need to figure out the alignment and sizing so it looks correct.

ronyeh:
See here for description of the FontFace API: https://developer.mozilla.org/en-US/docs/Web/API/CSS_Font_Loading_API

I believe it is equivalent to the `@font-face` blocks in CSS, except since it is in JS, we can control exactly when the fonts are loaded.

AaronDavidNewman:
> Can you clarify a bit? When you say put the woff in a CDN

I should have specified.  Maybe it will work but I think pushing an https script into javascript won't work in http context for security reasons.  But I could be wrong.

Anyway, I am getting this error:


```
Request URL:
https://unpkg.com/vexflow-fonts@1.0.5/bravura/Bravura_1.392.woff22
Request Method:
GET
Status Code:
404
Remote Address:
104.16.126.175:443
Referrer Policy:
strict-origin-when-cross-origin
```

AaronDavidNewman:
I fixed the error with the additional '2' (`  static async loadWebFont`) and I can now load the fonts.   But the Bravura text font is not getting used, it uses Times New Roman instead (which I can tell from chrome DOM inspector). 

Does this work for others?  I don't see how it would if Bravura doesn't have the ascii symbols, like Ron says.   No 'F7' b/c no 'F'.

rvilarl:
> I fixed the error with the additional '2' (` static async loadWebFont`) and I can now load the fonts. But the Bravura text font is not getting used, it uses Times New Roman instead (which I can tell from chrome DOM inspector).

Sorry for the 2, now updated.

> Does this work for others? I don't see how it would if Bravura doesn't have the ascii symbols, like Ron says. No 'F7' b/c no 'F'.

As @ronyeh  said Bravura / BravuraText combination makes no sense. I think that this is working fine for the Petaluma / PetalumanScript combination.

AaronDavidNewman:
> As @ronyeh said Bravura / BravuraText combination makes no sense

OK, then maybe we can move it back to Robot Slab for non-handwritten fonts.  I thought that worked the best with the chord changes, and it is a pretty complete font wrt different character sets, weights, styles etc.

AaronDavidNewman:
One more comment for the actual code part.  In your code, you space the blocks like so:

``` typescript
width += block.getWidth() + ChordSymbol.minPadding;
```
The old code had this:
``` typescript
block.width += ChordSymbol.spacingBetweenBlocks * fontSize * superSubScale;
```

This is why the chord changes are all squished together in the v5 version.  Since you put text and glyph in the same block, there is no spacing between them and so the symbols are both too small and too close to the text.  And the spacing needs to be scaled to the fonts size/super/sub so the spacing is the same in the superscripts.

V5:
image

vs.
image

AaronDavidNewman:
Another code comment: you left in the kerning tests, but removed the kerning logic.  

v5:
image

v4:
image

ronyeh:
@rvilarl For the purposes of this PR, can we go back to Roboto Slab just so we can get it as close as possible (visually) to VexFlow 4.x? It'll make things easier to reason about if we aren't changing too many things at the same time.

The big idea of this PR is how we draw & position glyphs, not that which text font we should pair with Bravura. In the future, we can revisit this and pick one of the following:
1. Have default VexFlow be completely web safe with regard to the text fonts.... so we would use boring fonts like Arial, Times New Roman, Georgia, for the ChordSymbols that are paired with the Bravura glyphs.
2. Use a web font and load it from CDN. Academico is the preferred text font as suggested by Steinberg designers. I'm not sure there is a woff2 available, as my copy of Dorico only has the OTFs. I think it is pretty easy to convert OTF => WOFF2 though.
3. Use a web font but install it via `npm install <font-name>`. For example, **fontsource** has Roboto Slab up on NPM: https://fontsource.org/fonts/roboto-slab/install  If we don't want to rely on **fontsource**'s npm packages, we could always roll our own npm packges (e.g., `npm install @vexflow-fonts/roboto-slab` or `npm install @vexflow-fonts/academico`).

Installing the font via npm install is an interesting idea, because you only need to be online to clone the vexflow repo and to do the NPM install. Afterward, you could work on a plane trip and your fonts should still show up correctly (as long as your dev server and bundler knows how to load the font).

Anyways, back to this PR. I vote for trying to minimize the visual differences as much as possible (so maybe we go back to Roboto Slab for now).

We should discuss kerning and other issues. I don't mind temporarily breaking some things, as long as we believe we have a possibility / path to restore those features before we ship 5.0 for real.

ronyeh:
> > I fixed the error with the additional '2' (` static async loadWebFont`) and I can now load the fonts. But the Bravura text font is not getting used, it uses Times New Roman instead (which I can tell from chrome DOM inspector).
> 
> Sorry for the 2, now updated.


Hi @rvilarl:

I wrote the original `loadWebFont()` function, and I do not believe it needs to be updated (to remove the "extra 2").

Here is the original version:

https://github.com/0xfe/vexflow/blob/master/src/font.ts#L350-L362

```ts
static async loadWebFont(fontName: string, woffURL: string, includeWoff2: boolean = true): Promise<FontFace> {
    const woff2URL = includeWoff2 ? `url(${woffURL}2) format('woff2'), ` : '';
    const woff1URL = `url(${woffURL}) format('woff')`;
    const woffURLs = woff2URL + woff1URL;
    // eslint-disable-next-line
    // @ts-ignore
    const fontFace = new FontFace(fontName, woffURLs);
    await fontFace.load();
    // eslint-disable-next-line
    // @ts-ignore
    document.fonts.add(fontFace);
    return fontFace;
}
```

The idea is that you pass in the woffURL, which points to a file named `my-awesome-font.woff`. By default, we will append the number `2` to the end of the *.woff URL, so we end up with `my-awesome-font.woff2`. Then we will provide BOTH woff2 and woff URLS to the `FontFace()` constructor. That's how web fonts are traditionally specified, with both a woff2 and a woff URL.

However, 96% of browsers support woff2 now, so if you really want to drop the support for the woff2 + woff approach that I wrote, I'm OK with that. We just need to make sure existing tests and demos don't break.

But I think you should be able to use my original `loadWebFont(...)` without modification. Just make sure you don't add the 2 at the end of the URLs you provide to the function.

src/font.ts
```diff
@@ -326,16 +324,34 @@ export class Font {
...
   static WEB_FONT_FILES: Record<string /* fontName */, string /* fontPath */> = {
-    'Roboto Slab': 'robotoslab/RobotoSlab-Medium_2.001.woff',
-    PetalumaScript: 'petaluma/PetalumaScript_1.10_FS.woff',
+    Bravura: 'bravura/Bravura_1.392.woff2',
+    BravuraText: 'bravura/BravuraText_1.393.woff2',
+    GonvilleSmufl: 'gonvillesmufl/GonvilleSmufl_1.100.woff2',
+    Gootville: 'gootville/Gootville_1.3.woff2',
+    GootvilleText: 'gootville/GootvilleText_1.2.woff2',
+    Leland: 'leland/Leland_0.75.woff2',
+    LelandText: 'leland/LelandText_0.75.woff2',
```
ronyeh, line 341: Note how the two deleted lines ended in *.woff.

There is no need to specify the woff2 file extension, because the loadWebFonts() function will automatically add the woff2 extension before providing it to the FontFace API.

src/flow.ts
```diff
@@ -228,8 +221,24 @@ export class Flow {
    * @returns an array of Font objects corresponding to the provided `fontNames`.
    */
   static setMusicFont(...fontNames: string[]): Font[] {
+    // HACK during review to introduce the correct stacks step by step
+    switch (fontNames[0]) {
+      case 'Bravura':
+        CommonMetrics.fontFamily = 'Bravura,BravuraText';
```
ronyeh, line 227: Yeah, we probably shouldn't use both Bravura and BravuraText, as they contain the same glyph code points.

src/flow.ts
```diff
@@ -228,8 +221,24 @@ export class Flow {
...
+        break;
+      case 'Gonville':
+        CommonMetrics.fontFamily = 'GonvilleSmufl,Bravura,Roboto Slab';
+        break;
+      case 'Leland':
+        CommonMetrics.fontFamily = 'Leland,Bravura,LelandText';
```
ronyeh, line 233: Same here. Drop LelandText from the font stack.

src/flow.ts
```diff
@@ -228,8 +221,24 @@ export class Flow {
    * @returns an array of Font objects corresponding to the provided `fontNames`.
    */
   static setMusicFont(...fontNames: string[]): Font[] {
+    // HACK during review to introduce the correct stacks step by step
+    switch (fontNames[0]) {
+      case 'Bravura':
+        CommonMetrics.fontFamily = 'Bravura,BravuraText';
+        break;
+      case 'Gonville':
+        CommonMetrics.fontFamily = 'GonvilleSmufl,Bravura,Roboto Slab';
```
ronyeh, line 230: This is interesting and unexpected for me. The font stack includes two music fonts and one text font.

I guess that sort of makes sense... if our ChordSymbol is a text string that includes a mix of music symbols and ASCII text symbols?

ronyeh (changes requested):
This is just a preliminary review. I haven't had the chance to read through every change yet.

I don't think loadWebFonts() needs to be modified at all. We just need to fix the URLs in WEB_FONT_FILES to not include the 2 at the end. Each file should end in `.woff`

ronyeh:
It looks like in Dorico, they use Bravura Text for chord symbols:

https://steinberg.help/dorico/v2/en/dorico/topics/notation_reference/notation_reference_chord_symbols_appearance_font_changing_t.html

> We recommend that you do not edit Chord Symbols Music Text Font, which is set to Bravura Text by default. This can only be set to a SMuFL-compliant font intended for use in text-based applications.

So I guess the app uses Bravura for the symbols on the staff, and Bravura Text for any symbols that happen in chord symbols, lyrics, and other places that place music symbols next to regular alpha numeric text.

ronyeh:
https://w3c.github.io/smufl/latest/tables/chord-symbols.html?highlight=chord#chord-symbols-ue870ue87f

At the bottom of the linked page:

> Scoring applications should be able to create strings with complex formatting, e.g. superscript and subscript characters, small digits stacked on top of each other, and scale these symbols to any arbitrary size in order to produce satisfactory chord symbols with a wide variety of visual appearances.

So it's the responsibility of the scoring application (Dorico... or in our case, VexFlow) to decide how to scale and kern the glyphs to make satisfactory looking chord symbols.

rvilarl:
@ronyeh @AaronDavidNewman please be patient, I am looking into option to improve the quality of the ChordSymbols taking into account the comments from both of you. Some notes:
- it is possible to make a single SVG Text that represent the complete symbol with sub/super indexes and different font sizes maiking use of tspan. However I do not see something similar for the canvas.
- the space between characters can be adjusted both in canvas and svg text with `letter-spacing`

@ronyeh could you create a vexflow-fonts with the text fonts suggested for Bravura, Leland, Gonville & Petaluma? One was Academico (for Bravura) and thanks for all your investigations on fonts!

rvilarl:
I have restored robot slab in all but petaluma, where I understood that the comibantion with PetalumaScript was fine.

tests/chordsymbol_tests.ts
```diff
@@ -69,7 +61,7 @@ function withModifiers(options: TestOptions): void {
     const formatter = f.Formatter();
     formatter.joinVoices([voice]);
     const voiceW = formatter.preCalculateMinTotalWidth([voice]);
-    const staffW = voiceW + Stave.defaultPadding + getGlyphWidth('gClef');
+    const staffW = voiceW + Stave.defaultPadding + getGlyphWidth('E050' /*gClef*/);
```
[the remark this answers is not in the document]
rvilarl: This is the approach. I have done it in src but I missed to look into tests.

src/font.ts
```diff
@@ -326,16 +324,34 @@ export class Font {
    * Alternative: https://cdn.jsdelivr.net/npm/vexflow-fonts@1.0.3/
    * Or you can use your own host.
    */
-  static WEB_FONT_HOST = 'https://unpkg.com/vexflow-fonts@1.0.3/';
+  static WEB_FONT_HOST = 'https://unpkg.com/vexflow-fonts@1.0.5/';
```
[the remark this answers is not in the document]
rvilarl: I agree. It is specially inconvenient whe you have no internet connection. I had to disable that several times flying.

src/font.ts
rvilarl: I decided to go explicitely to woff2 because there was a lot of discussion on network traffic and total length of files.

src/flow.ts
rvilarl: Roboto Slab already there. :)

src/flow.ts
rvilarl: It is indeed expected. The approach in Vexflow5 is to specify all the fonts and the browser will fall back to the next one when the glyph is not defined. We search in SMuFL first and then in standard text. My fault was to believe that BravuraText was a standard text file to combine with Bravura.

tests/font_tests.ts
```diff
@@ -53,17 +53,17 @@ function setFont(assert: Assert): void {
   const flat = new Accidental('b');
   // Add italic to the default font as defined in Element.TEXT_FONT (since Accidental does not override TEXT_FONT).
   flat.setFont(undefined, undefined, undefined, 'italic');
-  assert.equal(flat.getFont(), 'italic 30pt Bravura,Gonville,Custom');
+  assert.equal(flat.getFont(), 'italic 30pt Bravura,BravuraText');
```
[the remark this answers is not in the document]
rvilarl: Roboto Slab is now there.

src/font.ts
```diff
@@ -326,16 +324,34 @@ export class Font {
...
+    GonvilleSmufl: 'gonvillesmufl/GonvilleSmufl_1.100.woff2',
+    Gootville: 'gootville/Gootville_1.3.woff2',
+    GootvilleText: 'gootville/GootvilleText_1.2.woff2',
+    Leland: 'leland/Leland_0.75.woff2',
+    LelandText: 'leland/LelandText_0.75.woff2',
+    Petaluma: 'petaluma/Petaluma_1.065.woff2',
+    PetalumaScript: 'petaluma/PetalumaScript_1.10_FS.woff2',
+    MuseJazz: 'musejazz/MuseJazz_1.0.woff2',
+    MuseJazzText: 'musejazz/MuseJazzText_1.0.woff2',
+    'Roboto Slab': 'robotoslab/RobotoSlab-Medium_2.001.woff2',
```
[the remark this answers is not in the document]
rvilarl: changed

src/flow.ts
rvilarl: done

rvilarl:
I had to revert the renaming. If you install this google font on your system, it is register as 'Roboto Slab' and the images generated on the tests do not get the appropriated font. This was happening with 'Petaluma Script' as well.

rvilarl:
I did not find a way similar to tspan in canvas. The current bravura output is like this:

pptr-ChordSymbol Top_Chord_Symbols Bravura svg_current
pptr-ChordSymbol Bottom_Chord_Symbols Bravura svg_current
pptr-GlyphNote GlyphNote_with_ChordSymbols Bravura svg_current
pptr-ChordSymbol Top_Chord_Symbols_Justified Bravura svg_current
pptr-ChordSymbol Double_Bottom_Chord_Symbols Bravura svg_current
pptr-ChordSymbol Chord_Symbol_With_Modifiers Bravura svg_current
pptr-ChordSymbol Chord_Symbol_Kerning_Tests Bravura svg_current
pptr-ChordSymbol Chord_Symbol_Font_Size_Tests Bravura svg_current
pptr-ChordSymbol Bottom_Stem_Down_Chord_Symbols Bravura svg_current

rvilarl:
Petaluma outputs (here the kerning is not so important as A is a bit rotated):

pptr-ChordSymbol Top_Chord_Symbols Petaluma svg_current
pptr-ChordSymbol Bottom_Chord_Symbols Petaluma svg_current
pptr-GlyphNote GlyphNote_with_ChordSymbols Petaluma svg_current
pptr-ChordSymbol Top_Chord_Symbols_Justified Petaluma svg_current
pptr-ChordSymbol Double_Bottom_Chord_Symbols Petaluma svg_current
pptr-ChordSymbol Chord_Symbol_With_Modifiers Petaluma svg_current
pptr-ChordSymbol Chord_Symbol_Kerning_Tests Petaluma svg_current
pptr-ChordSymbol Chord_Symbol_Font_Size_Tests Petaluma svg_current
pptr-ChordSymbol Bottom_Stem_Down_Chord_Symbols Petaluma svg_current

rvilarl:
@AaronDavidNewman @mscuthbert @ronyeh what has to be fixed to make them acceptable?

ronyeh:
I think my main concern is if we do not have a (plausible) future solution for kerning, then we will be removing a feature that some folks (at least Aaron) like. :-)

If you think we have a possibility to reintroduce kerning for these symbols at some point (and can outline a rough idea as to how we might try to do it), then I would be okay with moving forward, even if it breaks kerning for now.

ronyeh:
In your most recent screenshots, I actually think the parentheses glyphs are too close to the other symbols. It feels too crowded to me. Are we over-kerning? In a way, if we removed kerning, everything would be slightly more spaced out and less crowded. Maybe that's good for some cases?

For example, in this screenshot you upoaded: https://user-images.githubusercontent.com/22865285/253409138-5363a582-faf8-4768-8a56-e12de89cbcd8.png

I think the # and b symbols are too close to the 11 and 9, respectively. I also think the parentheses are too close to the symbols as well, except for the one on the second row left side. That looks like kerning is turned off (and I actually kind of prefer the more spaced out look).

ronyeh:
> I had to revert the renaming. If you install this google font on your system, it is register as 'Roboto Slab' and the images generated on the tests do not get the appropriated font. This was happening with 'Petaluma Script' as well.

Ah yes... I agree it is a good idea to follow the original naming as defined by the font designer.

Just for the record.... do both 'Roboto Slab' has a space, and 'Petaluma Script' need a space? For some reason I thought PetalumaScript did NOT have a space inside.

We can verify by dragging font files onto this website to inspect the name: 

https://fontdrop.info/

Or you can also load it into the opentype.js website.

https://opentype.js.org/font-inspector.html


So... I think the confusing part is that there are multiple fields for a font's name.

If I inspect the Roboto Slab OTF, I get a windows naming table that says:

image

The postScriptName is `RobotoSlab-Medium` as PS does not allow spaces. The preferredFamily is `Roboto Slab`, with a space inside.


If I inspect the Petaluma Script OTF, I get:
```
fontFamily = Petaluma Script
fontSubfamily = Regular
fullName= PetalumaScript
postScriptName = PetalumaScript
```

So even though the postscriptName never has a space, I think we should be consistent and go with whatever is specified under `fontFamily`.

src/font.ts
```diff
@@ -347,13 +363,8 @@ export class Font {
...
   // @ts-ignore
-  static async loadWebFont(fontName: string, woffURL: string, includeWoff2: boolean = true): Promise<FontFace> {
-    const woff2URL = includeWoff2 ? `url(${woffURL}2) format('woff2'), ` : '';
-    const woff1URL = `url(${woffURL}) format('woff')`;
-    const woffURLs = woff2URL + woff1URL;
-    // eslint-disable-next-line
-    // @ts-ignore
-    const fontFace = new FontFace(fontName, woffURLs);
+  static async loadWebFont(fontName: string, woffURL: string): Promise<FontFace> {
+    const fontFace = new FontFace(fontName, `url(${woffURL})`);
```
ronyeh, line 367: If you look at the original version of loadWebFont, it specifies the woff2URL first, and then the woff1URL. The constructor accepts fallback fonts: https://stackoverflow.com/questions/41061599/how-to-load-fallback-font-formats-with-fontface-api

The browser will look for the woff2 file first, and if it returns a 404, it will fall back to the woff file. So this shouldn't have an impact on network traffic if the woff2 file exists and the user's browser supports woff2.

Anyways, I am OK if you change the API here since the original isn't the clearest. 

However, it would be nice if there was a way to support the fallback approach. The FontFace constructor does not require the url to point to a woff file. In fact, I've verified that you can pass in an OTF and my version of Chrome loads the OTF perfectly fine. :-)

Maybe the method signature here should be:

```
static async loadWebFont(fontName: string, fontURLs: string | string[]): Promise<FontFace> {
```

If we pass in a single string, then do what you are doing now. 

If we pass in a string[], then build up a comma separated list of fallback URLs?

rvilarl:
What about using the text font parenthesis?

pptr-ChordSymbol Chord_Symbol_With_Modifiers Bravura svg_current
pptr-ChordSymbol Chord_Symbol_With_Modifiers Petaluma svg_current

src/font.ts
rvilarl: What about providing the name without extension?

```
  static WEB_FONT_FILES: Record<string /* fontName */, string /* fontPath */> = {
    Bravura: 'bravura/Bravura_1.392',
    BravuraText: 'bravura/BravuraText_1.393',
...
 ```

and then adding them in the function.

**Comment by ronyeh** [comment]
Good idea. The only edge case that wouldn't solve is if the developer
prefers to load the OTF instead. We can always go with your approach for
now. Just rename it fontURL instead of woffURL. Later on we can accept a
string[] as well. We can do that in a separate PR.

**Comment by AaronDavidNewman** [comment]
> What about using the text font parenthesis?

How does it look?  'veryTall' parenthesis are supposed to be taller than 'tall' parenthesis.  They also have different placement (1/2 below the baseline).  I think those are what we used for stacked chords (x/y).

File: src/font.ts (hunk 6 above)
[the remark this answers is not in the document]
Comment (ronyeh) [reply src/font.ts:327]: I have been making npm packages for each individual font. Once installed via npm install, they should be available for local use in our project. Eventually, we will be able to support offline mode (e.g., coding on an airplane).

https://www.npmjs.com/settings/vexflow-fonts/packages

File: demos/node/fonts/load.js [hunk 10]
```diff
@@ -2,7 +2,7 @@
 //
 // This demo shows how to use node-canvas to load a font (woff / otf).
 // It does not currently use VexFlow.
-// TODO: Check if the text fonts (PetalumaScript & Roboto Slab) are automatically
+// TODO: Check if the text fonts (PetalumaScript & RobotoSlab) are automatically
```
ronyeh, line 5: Are we using spaces for both Petaluma Script and Roboto Slab?

src/font.ts
ronyeh: OK... I threw together (an incomplete and untested) approach for `loadWebFont()`

```ts
static async loadWebFont(fontFamily: string, fontURLs: string | string[], localNames?: string | string[]) {
    // Compose a src string that looks like:
    // local('FontName'), local('FontName-Bold'), url('font.woff2') format('woff2'), url('font.woff') format('woff')
    let sources: string[] = [];

    if (localNames) {
        if (typeof localNames === "string") {
            sources.push(`local('${localNames}')`);
        } else {
            for (const n of localNames) {
                sources.push(`local('${n}')`);
            }
        }
    }

    if (typeof fontURLs === "string") {
        fontURLs = [fontURLs];
    }

    for (const fontURL of fontURLs) {
        // Determine the type based on the extension.
        // .otf => opentype
        // .ttf => truetype
        // .woff => woff
        // .woff2 => woff2
        const extension = fontURL.split(".").pop();
        let fontType = "";
        switch (extension) {
            case "otf":
                fontType = "opentype";
                break;
            case "ttf":
                fontType = "truetype";
                break;
            case "woff":
                fontType = "woff";
                break;
            case "woff2":
            default:
                fontType = "woff2";
                break;
        }

        sources.push(`url('${fontURL}') format('${fontType}')`);
    }

    const src = sources.join(", ");
    console.log(src);
    // ... DO THE REST HERE
    // new FontFace(fontFamily, src).....
}

```

ronyeh: This new approach allows developers to load ONE or MORE fonts. The font types will be automatically detected by file extension.

Additionally, the developer can specify a local name (or more than one) so that the user agent can first look on the local system and load that font (without making a network request).

rvilarl:
> What about using the text font parenthesis?
> 
> pptr-ChordSymbol Chord_Symbol_With_Modifiers Bravura svg_current
> pptr-ChordSymbol Chord_Symbol_With_Modifiers Petaluma svg_current
> 
> 
> 

@AaronDavidNewman this is how the text parentheses look.

AaronDavidNewman:
They are short/fat compared to the stacked scenario.  For the ones on the same line, it's not too noticeable.

I realize I am comparing a screen shot to an svg, but I'm just focused on the the formatting:

Images from above:
image
image

VF 4:
image
image

ronyeh:
I guess it's a matter of personal preference.

For the V5 images that Aaron cropped, I do agree the Petaluma parentheses seem rather thick. However, the Roboto Slab parentheses match well with the Roboto Slab font.

In the V4 images, I think the Bravura parentheses do not visually pair well with the Roboto Slab font. The Bravura parentheses are too thin in my opinion. Also, I think the 7 in the F7 is too close to the left parentheses in the last image (Roboto + Bravura).

I am fine with either solution.

In theory I would prefer the parentheses to be from the **text font**, because to me, it flows better visually (especially in the Roboto Slab example).

However, SMuFL defines a bunch of parentheses glyphs for chord symbols, so we might as well use them :-)
https://w3c.github.io/smufl/latest/tables/chord-symbols.html

I think in this case, Academico might fit better alongside Bravura, since the Roboto Slab variant that we are using has a heavier weight than the Bravura parentheses symbols.


Regarding this PR, I would lean towards keeping it as close as possible to the V4 look and feel.

In the interest of moving forward, I'd be okay if this PR did not include a solution for kerning, as long as we keep the issue open and can revisit it before shipping 5.0. We can always learn from the approach that Aaron took and see if we can bring some of those ideas over to 5.0.

src/flow.ts
```diff
@@ -228,8 +221,24 @@ export class Flow {
    * @returns an array of Font objects corresponding to the provided `fontNames`.
    */
   static setMusicFont(...fontNames: string[]): Font[] {
+    // HACK during review to introduce the correct stacks step by step
```
ronyeh, line 224: So after this PR is merged, will this "HACK" go away? What is the correct solution in this method?

tools/fonts/inspector/otf.html
```diff
@@ -214,7 +214,7 @@
         await gonville.load();
         document.fonts.add(gonville);
         const robotoSlabOTF = '../@/robotoslab/RobotoSlab-Medium_2.001.otf';
-        const robotoSlab = new FontFace('Roboto Slab', `url(${robotoSlabOTF}) format('opentype')`);
+        const robotoSlab = new FontFace('RobotoSlab', `url(${robotoSlabOTF}) format('opentype')`);
```
ronyeh, line 217: Just checking if we are using spaces in all the font names. Is it supposed to be "Roboto Slab", with a space?

AaronDavidNewman:
> In the interest of moving forward, I'd be okay if this PR did not include a solution for kerning

Yes, we should definitely move on.

I wonder now if it even makes sense to keep the chord symbols module.  The thing it offered in V4 is 1) the ability to mix text and glyphs, and 2) the ability to pass in your own metrics outside of VexFlow main build.  1 is no longer required, because we are using the glyphs/test directly from the font.  And 2 is no longer possible because there are no longer any metrics to use for formatting.

So the only thing in ChordSymbol that isn't also in annotation is super/subscript, which we could move to annotation at some future time.  It might be useful for other things.

If you want to keep it, you can probably just go with what you have.

ronyeh:
I'd vote to keep the chord symbols module for now. It makes it easier to
reason about our migration from v4 to v5, as we can compare features
module-to-module.

At some point once things are more stable, we might realize that
chordsymbol is largely redundant.... Then maybe it could be refactored as a
subclass or eliminated/merged into another class.

src/flow.ts
rvilarl: The hack will stay until all the code is ported this hack avoids that the non ported files produce visual diffrences. In the end it will be:

```
static setMusicFont(...fontNames: string[]): void {
    // Convert the array of font names into an array of Font objects.
    CommonMetrics.fontFamily = fontNames.join(',');
  }
```

src/font.ts
rvilarl: I suggest tha you make a PR for this one

demos/node/fonts/load.js
rvilarl: done

tools/fonts/inspector/otf.html
rvilarl: done

tests/font_tests.ts
```diff
@@ -75,7 +74,7 @@ function fontParsing(assert: Assert): void {
   assert.equal(bFont?.weight, FontWeight.NORMAL);
   assert.equal(bFont?.style, FontStyle.NORMAL);
 
-  const f1 = 'Roboto Slab, serif';
+  const f1 = 'RobotoSlab, serif';
```
ronyeh, line 77: Should it be 'Roboto Slab', with a space?

src/chordsymbol.ts
```diff
@@ -628,33 +351,12 @@ export class ChordSymbol extends Modifier {
...
-
-  setEnableKerning(val: boolean): this {
-    this.useKerning = val;
-    return this;
+  addLine(
+    params: Partial<{
+      symbolModifier: SymbolModifiers;
+    }> = {}
+  ): this {
+    return this.addText(String.fromCharCode(0xe874, 0xe874), params);
```
ronyeh, line 359: I'm beginning to think that we need comments for each SMuFL unicode codepoint that we stick in our code base. 

`0xe874` means nothing to me :-)

Maybe we should always add a comment above lines that include SMuFL code points?

```ts
// SMuFL U+E874 / csymMinor / Minor
return this.addText(String.fromCharCode(0xe874, 0xe874), params);
```

If you do it for this one, I can help out with other ones by submitting PRs. I think having this comment will help future readers of our code.

tests/font_tests.ts
rvilarl: done

src/chordsymbol.ts
rvilarl: Changed like other comments in the file with:
/*csymMinor*/

rvilarl:
Is there anything still to do in this one?

ronyeh:
I'll give it one more look and then will merge if there's nothing major. Probably by Tuesday your time.

src/chordsymbol.ts
```diff
@@ -9,51 +9,42 @@
...
-    superSubRatio: number;
-  };
-  glyphs: Record<string, ChordSymbolGlyphMetrics>;
+  isSuperscript(): boolean {
+    return this.symbolModifier !== undefined && this.symbolModifier === SymbolModifiers.SUPERSCRIPT;
+  }
+
+  isSubscript(): boolean {
+    return this.symbolModifier !== undefined && this.symbolModifier === SymbolModifiers.SUBSCRIPT;
+  }
```
ronyeh, line 47: In both of these functions, is the undefined check unnecessary?

Can't we just check if the this.symbolModifier === SUPERSCRIPT or SUBSCRIPT? Not sure why we need to check for undefined.

src/chordsymbol.ts
```diff
@@ -111,166 +96,49 @@ export class ChordSymbol extends Modifier {
...
-  static isSubscript(block: ChordSymbolBlock): boolean {
-    return block.symbolModifier !== undefined && block.symbolModifier === SymbolModifiers.SUBSCRIPT;
-  }
-
   static get minPadding(): number {
-    const musicFont = Tables.currentMusicFont();
-    return musicFont.lookupMetric('noteHead.minPadding');
+    return Tables.lookupMetric('NoteHead.minPadding');
   }
   /**
```
ronyeh, line 143: Do we need an extra empty line here?

src/chordsymbol.ts
```diff
@@ -280,108 +148,67 @@ export class ChordSymbol extends Modifier {
...
-        if (
-          block.symbolType === SymbolTypes.GLYPH &&
-          block.glyph !== undefined &&
-          block.glyph.code === ChordSymbol.glyphs.over.code
-        ) {
-          lineSpaces = 2;
-        }
-        block.width += ChordSymbol.spacingBetweenBlocks * fontSize * superSubScale;
-
         // If a subscript immediately  follows a superscript block, try to
```
ronyeh, line 172: This line has an extra space in `"immediately  follows"`. I know it might not be your comment, but we might as well fix it since we're here....

src/chordsymbol.ts
```diff
@@ -395,227 +222,123 @@ export class ChordSymbol extends Modifier {
...
-
-    this.textFormatter = TextFormatter.create(this.textFont);
   }
 
   /**
    * The offset is specified in `em`. Scale this value by the font size in pixels.
    */
   get superscriptOffset(): number {
-    return ChordSymbol.superscriptOffset * this.textFormatter.fontSizeInPixels;
+    return ChordSymbol.superscriptOffset * Font.convertSizeToPixelValue(this.textFont?.size);
```
ronyeh, line 237: I see you do this twice:
```
this.textFont?.size
```

Can textFont be undefined? Why are we doing the optional chaining `?.` operator here?

If it is possible for it to be undefined, it will pass undefined to `Font.convertSizeToPixelValue(...)`. I don't believe that function can accept an undefined parameter.

Can we remove the ? from the two places we use it? I thought textFont is required to be defined.

src/chordsymbol.ts
```diff
@@ -395,227 +222,123 @@ export class ChordSymbol extends Modifier {
...
-    if (preKernLower && currSymbol.symbolType === SymbolTypes.TEXT) {
-      if (currSymbol.text[0] >= 'a' && currSymbol.text[0] <= 'z') {
-        adjustment += kerningOffsetPixels / 2;
-      }
-      if (ChordSymbol.upperKerningText.some((xx) => xx === prevSymbol?.text[prevSymbol.text.length - 1])) {
-        adjustment += kerningOffsetPixels / 2;
-      }
-    }
-    return adjustment;
+    return ChordSymbol.subscriptOffset * Font.convertSizeToPixelValue(this.textFont?.size);
```
ronyeh, line 241: See the other comment about optional chaining. Let's make sure we aren't accidentally passing `undefined` to Font.convertSizeToPixelValue().

src/chordsymbol.ts
```diff
@@ -395,227 +222,123 @@ export class ChordSymbol extends Modifier {
...
-    }
-
-    for (symIndex = barIndex + 1; symIndex < this.symbolBlocks.length; ++symIndex) {
-      const symbol = this.symbolBlocks[symIndex];
-      symbol.xShift = symbol.xShift - xoff;
-      symbol.yShift = symbol.yShift + yoff;
-    }
-  }
-
-  updateKerningAdjustments(): void {
```
ronyeh, line 466: I'm OK with removing the kerning code for now. If we ever decide we need it in V5, we can come back to this PR and see if the V4 approach can help us.

src/chordsymbol.ts
```diff
@@ -395,227 +222,123 @@ export class ChordSymbol extends Modifier {
...
+    const symbolBlock = new ChordSymbolBlock(
+      params.text ?? '',
+      params.symbolModifier ?? SymbolModifiers.NONE,
+      0,
+      0,
+      false
+    );
+
+    if (symbolBlock.isSubscript()) symbolBlock.setYShift(this.subscriptOffset);
+    if (symbolBlock.isSuperscript()) symbolBlock.setYShift(this.superscriptOffset);
```
ronyeh, line 263: In general I am not a fan of single line if statements. I'd prefer to always use braces even if it's only one statement:

```js
if (condition) {
    <a single line of code>
}
```

It sets us up for the future and prevents possible bugs (in the future).

tests/chordsymbol_tests.ts
```diff
@@ -39,18 +38,11 @@ const note = (factory: Factory, keys: string[], duration: string, chordSymbol: C
...
-  //   NOTATION_FONT_SCALE: 39,
-  //   TABLATURE_FONT_SCALE: 39,
-  const musicFont = Tables.currentMusicFont();
-  const glyph: FontGlyph = musicFont.getGlyphs()[glyphName];
-  const widthInEm = (glyph.xMax - glyph.xMin) / musicFont.getResolution();
-  return widthInEm * 38 * Font.scaleToPxFrom.pt;
+function getGlyphWidth(charCode: number): number {
+  const el = new Element();
+  el.setText(String.fromCharCode(charCode));
+  el.measureText()
```
ronyeh, line 44: We can discuss this after this PR is merged. Should setText include `this.measureText()` as the last line of its implementation? That way, every time you setText(), the width will be valid. Is there a big performance impact if we do this?

I'm just worried since every time we call setText(...) and change the text, the width is now incorrect. A future developer (or ourselves someday, when we forget this detail), might not know that the width is now incorrect, and can only be fixed by calling measureText() again.

ronyeh (changes requested):
I'm done reviewing, and most of my comments suggest minor changes.

I'll merge it once those are addressed. Thanks!

src/chordsymbol.ts
rvilarl: you are right, done

src/chordsymbol.ts
rvilarl: done

src/chordsymbol.ts
rvilarl: done

src/chordsymbol.ts
rvilarl: done

src/chordsymbol.ts
rvilarl: ok

src/chordsymbol.ts
rvilarl: done

tests/chordsymbol_tests.ts
rvilarl: There are elements adding padding. I agree that we should look into that. Perhaps an issue to resolve after the migration is complete?

ronyeh:
Thanks!

PR #16, opened 2023-07-10
Update repo URLs from `0xfe/vexflow` to `vexflow/vexflow` to make the release script push to the correct repository.

ronyeh:
The most important changes are in Gruntfile.js and package.json. For example:

```
pushRepo: 'git@github.com:vexflow/vexflow.git',
```

This will allow our release script and NPM to reference the correct VexFlow 5+ repo.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -386,7 +386,7 @@ module.exports = (grunt) => {
       files: ['tests/flow-headless-browser.html'],
       options: {
         puppeteer: {
-          headless: "new",
+          headless: 'new',
           args: ['--no-sandbox', '--disable-setuid-sandbox'],
         },
       },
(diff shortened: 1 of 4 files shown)


Gruntfile.js
```diff
@@ -848,7 +848,7 @@ module.exports = (grunt) => {
 
       const message =
         '\nconsole.warn("Please upgrade to the newest release of VexFlow.\\n' +
-        'See: https://github.com/0xfe/vexflow for more information.\\nThis page uses version 3.0.9, which is no longer supported.");\n\n' +
+        'See: https://github.com/vexflow/vexflow for more information.\\nThis page uses version 3.0.9, which is no longer supported.");\n\n' +
```
rvilarl, line 851: Does the 3.0.9 message make sense?

rvilarl (changes requested):

ronyeh:
Once your V5 work is all done, we can go through old stuff (e.g. deprecated
or irrelevant code) and start deleting it.

This 3.0.9 message is targeted at old webpages that linked directly to the
unpkg release. For example, jsfiddle examples.

If we released the 4.x or 5.x version at that same URL, all those old pages
would immediately break.

So when we released 4.0, we put it at a different path and in the readme,
we advise people to include the version number in their script tag so
future releases won't break their pages.

Every time we ship a new release, we also ship 3.0.9 with an extra message
appended to the end of that script saying it is now outdated.

At some point we can decide it is okay to break these old pages. But for
now we have been keeping them alive.

rvilarl:
ok

Issue #14, opened 2023-07-10
Should the CommonMetrics class be in its own file? 


rvilarl:
To be decided once vexflow5 alpha 0 is reviewed.

ronyeh:
First discussed here:

rvilarl:
I know you care a bit :-) about how big the final library will be. This is obviously premature optimization, but later on, if we discover that our minimizer leaves all these strings intact, it might be fun to have a local variable internal to a common_metrics.ts file that declares `'Times New Roman, serif'` and `'Arial, sans-serif'` so that CommonMetrics can just reference those internal variables.

_Originally posted by @ronyeh in https://github.com/vexflow/vexflow/pull/18#discussion_r1258831967_

rvilarl:
This class could also have the lookupmetrics member function and make it available to the users, removing the need to import tables on tests.

rvilarl:
Followed up in #78

PR #18, opened 2023-07-10
element (v5)

### LINKED ISSUE #17: Comments on #12
Hi Rodrigo!

I wasn't finished reviewing [PR #12](https://github.com/vexflow/vexflow/pull/12). I needed more time to review each file.

Sorry, I should have been more clear. In the future, could you wait for a "Looks Good To Me / LGTM" before merging? Or you could provide your feedback or address the comments I made and then I can go ahead and merge.

I understand that you have been working on your approach for a while now, but to the rest of us who are also interested in VexFlow, some of the concepts and the specific approach you took are completely new information. I personally trust that you made good design decisions, but we all need just a bit more time to process this new information (and ideally, to read through each line of code you wrote).

You're a maintainer of VexFlow 5, so of course you can go ahead and merge any and all PRs you wish. But if you don't give folks like Michael, Mohit, myself, Aaron, Simon, Greg, etc a chance to comment on or review your work, VexFlow 5 might ultimately end up being "Rodrigo's branch" and everyone else might hesitate to upgrade from VexFlow 4.x.

_Originally posted by @ronyeh in https://github.com/vexflow/vexflow/issues/12#issuecomment-1629342448_

rvilarl:
Fixes #17 
Once again in review, sorry for the misunderstanding. Previous review comments in #12 already pushed.

### DIFF
--- a/src/chordsymbol.ts
+++ b/src/chordsymbol.ts
@@ -9,7 +9,7 @@
 //
 // See `tests/chordsymbol_tests.ts` for usage examples.
 
-import { Font, FontInfo, FontStyle, FontWeight } from './font';
+import { Font, FontInfo } from './font';
 import { Glyph } from './glyph';
 import { Modifier } from './modifier';
 import { ModifierContextState } from './modifiercontext';
@@ -409,26 +409,12 @@ export class ChordSymbol extends Modifier {
 
   constructor() {
     super();
-    this.resetFont();
-  }
-
-  /**
-   * Default text font.
-   * Choose a font family that works well with the current music engraving font.
-   * @override `Element.TEXT_FONT`.
-   */
-  static get TEXT_FONT(): Required<FontInfo> {
-    let family = 'Roboto Slab, Times, serif';
     if (Tables.currentMusicFont().getName() === 'Petaluma') {
       // Fixes Issue #1180
-      family = 'PetalumaScript, Arial, sans-serif';
+      this.textFont.family = 'PetalumaScript, Arial, sans-serif';
     }
-    return {
-      family,
-      size: 12,
-      weight: FontWeight.NORMAL,
-      style: FontStyle.NORMAL,
-    };
+
+    this.textFormatter = TextFormatter.create(this.textFont);
   }
 
   /**
--- a/src/element.ts
+++ b/src/element.ts
@@ -3,11 +3,12 @@
 // @author Mohit Cheppudira
 
 import { BoundingBox } from './boundingbox';
-import { Font, FontInfo, FontStyle, FontWeight } from './font';
+import { Font, FontInfo } from './font';
 import { Registry } from './registry';
 import { RenderContext } from './rendercontext';
+import { Tables } from './tables';
 import { Category } from './typeguard';
-import { defined, prefix } from './util';
+import { defined, prefix, RuntimeError } from './util';
 
 /** Element attributes. */
 export interface ElementAttributes {
@@ -59,12 +60,11 @@ export interface ElementStyle {
  * Element implements a generic base class for VexFlow, with implementations
  * of general functions and properties that can be inherited by all VexFlow elements.
  *
- * The Element is an abstract class that needs to be subclassed to work. It handles
- * style and text-font properties for the Element and any child elements, along with
- * working with the Registry to create unique ids, but does not have any tools for
- * formatting x or y positions or connections to a Stave.
+ * The Element handles style and text-font properties for the Element and any child
+ * elements, along with working with the Registry to create unique ids, but does not
+ * have any tools for formatting x or y positions or connections to a Stave.
  */
-export abstract class Element {
+export class Element {
   static get CATEGORY(): string {
     return Category.Element;
   }
@@ -77,17 +77,6 @@ export abstract class Element {
     return `auto${Element.ID++}`;
   }
 
-  /**
-   * Default font for text. This is not related to music engraving. Instead, see `Flow.setMusicFont(...fontNames)`
-   * to customize the font for musical symbols placed on the score.
-   */
-  static TEXT_FONT: Required<FontInfo> = {
-    family: Font.SANS_SERIF,
-    size: Font.SIZE,
-    weight: FontWeight.NORMAL,
-    style: FontStyle.NORMAL,
-  };
-
   #context?: RenderContext;
   #attrs: ElementAttributes;
 
@@ -97,20 +86,38 @@ export abstract class Element {
   protected registry?: Registry;
 
   /**
-   * Some elements include text.
    * The `textFont` property contains information required to style the text (i.e., font family, size, weight, and style).
-   * It is undefined by default, and can be set using `setFont(...)` or `resetFont()`.
    */
-  protected textFont?: Required<FontInfo>;
+  protected textFont: Required<FontInfo>;
+  protected text = '';
+  protected textMetrics: TextMetrics = {
+    fontBoundingBoxAscent: 0,
+    fontBoundingBoxDescent: 0,
+    actualBoundingBoxAscent: 0,
+    actualBoundingBoxDescent: 0,
+    actualBoundingBoxLeft: 0,
+    actualBoundingBoxRight: 0,
+    width: 0,
+  };
 
-  constructor() {
+  protected width: number = 0;
+  protected yShift: number = 0;
+  protected xShift: number = 0;
+
+  constructor(category?: string) {
     this.#attrs = {
       id: Element.newID(),
-      type: this.getCategory(),
+      type: category ? category : (<typeof Element>this.constructor).CATEGORY,
       class: '',
     };
 
     this.rendered = false;
+    this.textFont = {
+      family: Tables.lookupMetric(`${this.#attrs.type}.fontFamily`),
+      size: Tables.lookupMetric(`${this.#attrs.type}.fontSize`),
+      weight: Tables.lookupMetric(`${this.#attrs.type}.fontWeight`),
+      style: Tables.lookupMetric(`${this.#attrs.type}.fontStyle`),
+    };
 
     // If a default registry exist, then register with it right away.
     Registry.getDefaultRegistry()?.register(this);
@@ -131,7 +138,7 @@ export abstract class Element {
   }
 
   getCategory(): string {
-    return (<typeof Element>this.constructor).CATEGORY;
+    return `${this.#attrs.type}`;
   }
 
   /**
@@ -213,8 +220,9 @@ export abstract class Element {
   }
 
   /** Draw an element. */
-  // eslint-disable-next-line
-  abstract draw(...args: any[]): void;
+  draw(): void {
+    throw new RuntimeError('Element', 'Draw not defined');
+  }
 
   /** Check if it has a class label (An element can have multiple class labels).  */
   hasClass(className: string): boolean {
@@ -351,12 +359,15 @@ export abstract class Element {
    * Each Element subclass may specify its own default by overriding the static `TEXT_FONT` property.
    */
   setFont(font?: string | FontInfo, size?: string | number, weight?: string | number, style?: string): this {
-    // Allow subclasses to override `TEXT_FONT`.
-    const defaultTextFont: Required<FontInfo> = (<typeof Element>this.constructor).TEXT_FONT;
+    const defaultTextFont: Required<FontInfo> = {
+      family: Tables.lookupMetric(`${this.#attrs.type}.fontFamily`),
+      size: Tables.lookupMetric(`${this.#attrs.type}.fontSize`),
+      weight: Tables.lookupMetric(`${this.#attrs.type}.fontWeight`),
+      style: Tables.lookupMetric(`${this.#attrs.type}.fontStyle`),
+    };
 
     const fontIsObject = typeof font === 'object';
     const fontIsString = typeof font === 'string';
-    const fontIsUndefined = font === undefined;
     const sizeWeightStyleAreUndefined = size === undefined && weight === undefined && style === undefined;
 
     if (fontIsObject) {
@@ -365,13 +376,7 @@ export abstract class Element {
     } else if (fontIsString && sizeWeightStyleAreUndefined) {
       // `font` is case 2) CSS font shorthand.
       this.textFont = Font.fromCSSString(font);
-    } else if (fontIsUndefined && sizeWeightStyleAreUndefined) {
-      // All arguments are undefined. Do not check for `arguments.length === 0`,
-      // which fails on the edge case: `setFont(undefined)`.
-      // TODO: See if we can remove this case entirely without introducing a visual diff.
-      // The else case below seems like it should be equivalent to this case.
-      this.textFont = { ...defaultTextFont };
-    } else {
+     } else {
       // `font` is case 3) a font family string (e.g., 'Times New Roman').
       // The other parameters represent the size, weight, and style.
       // It is okay for `font` to be undefined while one or more of the other arguments is provided.
@@ -383,6 +388,7 @@ export abstract class Element {
         style ?? defaultTextFont.style
       );
     }
+    this.measureText();
     return this;
   }
 
@@ -391,28 +397,14 @@ export abstract class Element {
    * 'bold 10pt Arial'.
    */
   getFont(): string {
-    if (!this.textFont) {
-      this.resetFont();
-    }
     return Font.toCSSString(this.textFont);
-  }
-
-  /**
-   * Reset the text font to the style indicated by the static `TEXT_FONT` property.
-   * Subclasses can call this to initialize `textFont` for the first time.
-   */
-  resetFont(): void {
-    this.setFont();
   }
 
   /** Return a copy of the current FontInfo object. */
   get fontInfo(): Required<FontInfo> {
-    if (!this.textFont) {
-      this.resetFont();
-    }
     // We can cast to Required<FontInfo> here, because
     // we just called resetFont() above to ensure this.textFont is set.
-    return { ...this.textFont } as Required<FontInfo>;
+    return this.textFont;
   }
 
   set fontInfo(fontInfo: FontInfo) {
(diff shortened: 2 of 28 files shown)


ronyeh:
Thanks! Sorry for the confusion. I don't have much time in the evening to do code review work, so sometimes it might have to be split up over several days.

Once I'm done reviewing, I will go ahead and merge.

I think it's probably also slightly better if we don't merge our own PRs, haha. For example, you can review and merge my PRs, while I can review and merge your PRs.... That way, a casual observer to this repo might be more confident that all the code has at least been seen by two people.

src/strokes.ts
```diff
@@ -243,8 +233,14 @@ export class Stroke extends Modifier {
 
     // Draw the rasquedo "R"
     if (this.type === Stroke.Type.RASQUEDO_DOWN || this.type === Stroke.Type.RASQUEDO_UP) {
+      const textFont = {
+        family: Tables.lookupMetric(`Strokes.text.fontFamily`),
+        size: Tables.lookupMetric(`Strokes.text.fontSize`),
+        weight: Tables.lookupMetric(`Strokes.text.fontWeight`),
+        style: Tables.lookupMetric(`Strokes.text.fontStyle`),
+      };
```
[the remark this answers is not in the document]
ronyeh: If the `CommonMetrics` was defined with a `font` key, perhaps we could do this?

For example:
```js
FretHandFinger: {
  font: {
    family: 'Arial, sans-serif',
    size: 9,
    weight: 'bold', 
  },
},

PedalMarking: {
  font: {
    family: 'Times New Roman, serif',
    size: 12,
    weight: 'bold',
    style: 'italic',
  },
},
```

Then maybe we can do a lookupMetric('PedalMarking.font') to retrive the entire object?

src/element.ts
```diff
@@ -491,4 +488,95 @@ export abstract class Element {
...
+    return this;
+  }
+
+  /** Get element text */
+  getText(): string {
+    return this.text;
+  }
+
+  /**
+   * Render the element text .
```
ronyeh, line 541: There is an extra space before the period.

src/element.ts
```diff
@@ -491,4 +488,95 @@ export abstract class Element {
...
+    this.text = text;
+    return this;
+  }
+
+  /** Get element text */
+  getText(): string {
+    return this.text;
+  }
+
+  /**
```
ronyeh, line 540: Inconsistent style of one-line comments. I'm OK with either as long as we are more locally consistent. :-)

In fact, I'm totally OK with the more verbose form, even for single line comments.

```js
/**
 * Here is a single-line comment.
 */
```

This allows us to easily expand the comment in the future if necessary.

Also.... this is extremely nitpicky, but I think comments that are complete sentences should end with the correct punctuation, like a period or question mark if it's a question. It just makes the code base look nicer overall.

src/stavetie.ts
```diff
@@ -168,7 +160,7 @@ export class StaveTie extends Element {
    * @param firstX specified in pixels
    * @param lastX specified in pixels
    */
-  renderText(firstX: number, lastX: number): void {
+  renderTieText(firstX: number, lastX: number): void {
     if (!this.text) return;
```
ronyeh, line 164: Is this line still necessary? Can this.text be undefined? I noticed that you changed the constructor to default the text to '' instead of undefined.

src/staverepetition.ts
```diff
@@ -13,13 +13,6 @@ export class Repetition extends StaveModifier {
     return Category.Repetition;
   }
 
-  static TEXT_FONT: Required<FontInfo> = {
-    family: Font.SERIF,
-    size: Tables.NOTATION_FONT_SCALE / 3,
```
ronyeh, line 18: It used to be 1/3 of the NOTATION_FONT_SCALE. So I assume you just hardcode a number now in the CommonMetrics?

src/flow.ts
```diff
@@ -186,7 +180,6 @@ export class Flow {
   static AnnotationVerticalJustify = AnnotationVerticalJustify;
   static ChordSymbolHorizontalJustify = ChordSymbolHorizontalJustify;
   static ChordSymbolVerticalJustify = ChordSymbolVerticalJustify;
-  static SymbolTypes = SymbolTypes;
```
ronyeh, line 189: I assume this deletion is related to the next PR about ChordSymbol?

src/element.ts
```diff
@@ -491,4 +488,95 @@ export abstract class Element {
...
+
+  /** Canvas used to measure text */
+  static #txtCanvas?: HTMLCanvasElement;
+
+  measureText(): TextMetrics {
+    let txtCanvas = Element.#txtCanvas;
+    if (!txtCanvas) {
+      // Create the SVG text element that will be used to measure text in the event
+      // of a cache miss.
+      txtCanvas = document.createElement('canvas');
```
ronyeh, line 558: Is the comment incorrect? Here you create a `canvas` element.

Will this approach work for all the environments that VexFlow 4 supports? In this case, we require the environment to be able to create a HTMLCanvasElement. What about NodeJS environments or [web workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API)?

I know some other folks added support for [OffscreenCanvas](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas) to VexFlow, and there was node canvas.

Anyways, it is OK for VexFlow 5 to (temporarily) break support for those environments, but we should we aware that this approach (**requiring a HTMLCanvasElement for measurements**) is not a solution for everyone.

src/element.ts
```diff
@@ -491,4 +488,95 @@ export abstract class Element {
...
+      Element.#txtCanvas = txtCanvas;
+    }
+    const context = txtCanvas.getContext('2d');
+    if (!context) throw new RuntimeError('Font', 'No txt context');
+    context.font = Font.toCSSString(Font.validate(this.textFont));
+    this.textMetrics = context.measureText(this.text);
```
ronyeh, line 564: If we decide to verify support for `OffscreenCanvas`... we should check that measureText() works here as we expect it to.

https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvasRenderingContext2D

src/element.ts
```diff
@@ -491,4 +488,95 @@ export abstract class Element {
...
+    this.boundingBox = new BoundingBox(
+      0,
+      -this.textMetrics.actualBoundingBoxAscent,
+      this.textMetrics.width,
+      this.textMetrics.actualBoundingBoxDescent + this.textMetrics.actualBoundingBoxAscent
```
ronyeh, line 569: We could extract `this.textMetrics.actualBoundingBoxAscent` => `const ascent = this.textMetrics.actualBoundingBoxAscent`

src/element.ts
```diff
@@ -97,20 +92,37 @@ export abstract class Element {
...
+    actualBoundingBoxAscent: 0,
+    actualBoundingBoxDescent: 0,
+    actualBoundingBoxLeft: 0,
+    actualBoundingBoxRight: 0,
+    width: 0,
+  };
+
+  protected width: number = 0;
+  protected yShift: number = 0;
+  protected xShift: number = 0;
```
ronyeh, line 110: Should the `xShift` line be on top? I normally think of x, y...

src/element.ts
```diff
@@ -97,20 +92,37 @@ export abstract class Element {
...
       class: '',
     };
 
     this.rendered = false;
+    this.textFont = {
+      family: Tables.lookupMetric(`${this.#attrs.type}.fontFamily`),
+      size: Tables.lookupMetric(`${this.#attrs.type}.fontSize`),
+      weight: Tables.lookupMetric(`${this.#attrs.type}.fontWeight`),
+      style: Tables.lookupMetric(`${this.#attrs.type}.fontStyle`),
+    };
```
ronyeh, line 125: It would be nice if this could be handled on a single line. Something like:

```js
this.textFont = Tables.lookupMetric(this.#attrs.type.font);
```

src/element.ts
```diff
@@ -351,12 +364,15 @@ export abstract class Element {
...
    */
   setFont(font?: string | FontInfo, size?: string | number, weight?: string | number, style?: string): this {
-    // Allow subclasses to override `TEXT_FONT`.
-    const defaultTextFont: Required<FontInfo> = (<typeof Element>this.constructor).TEXT_FONT;
+    const defaultTextFont: Required<FontInfo> = {
+      family: Tables.lookupMetric(`${this.#attrs.type}.fontFamily`),
+      size: Tables.lookupMetric(`${this.#attrs.type}.fontSize`),
+      weight: Tables.lookupMetric(`${this.#attrs.type}.fontWeight`),
+      style: Tables.lookupMetric(`${this.#attrs.type}.fontStyle`),
+    };
```
ronyeh, line 372: It would be nice if this could be handled on a single line. Something like:

```js
this.textFont = Tables.lookupMetric(this.#attrs.type.font);
```

src/element.ts
```diff
@@ -213,8 +225,9 @@ export abstract class Element {
   }
 
   /** Draw an element. */
-  // eslint-disable-next-line
-  abstract draw(...args: any[]): void;
+  draw(): void {
+    throw new RuntimeError('Element', 'Draw not defined');
```
ronyeh, line 229: Since Element is no longer abstract, it would be nice if the methods didn't throw RuntimeErrors.

Maybe draw() needs some default implementation to draw an empty string at xShift=0, yShift=0?

src/element.ts
```diff
@@ -59,12 +60,17 @@ export interface ElementStyle {
...
- * style and text-font properties for the Element and any child elements, along with
- * working with the Registry to create unique ids, but does not have any tools for
- * formatting x or y positions or connections to a Stave.
+ * The Element handles style and font properties for the Element and any child
+ * elements, along with working with the Registry to create unique ids.
+ *
+ * The `text` is a series of unicode characters (including SMuFL codes).
+ * The `textFont` property contains information required to style the text (i.e., font family, size, weight, and style). 
+ * This font family is a comma separated list of fonts. 
+ * The method `measureText` calculates the `textMetrics` and `ẃidth` of the `text`.
```
ronyeh, line 69: There is a typo in `width`. There seems to be an accent / accute mark on the w.

src/element.ts
```diff
@@ -59,12 +60,17 @@ export interface ElementStyle {
...
+ * The methos `renderText` will render the text using the context provided at the coordinates provided
```
ronyeh, line 70: Typo: methos => method

You can use the following:
```
The method `renderText(...)` will render the text using the provided context and coordinates.
```

src/chordsymbol.ts
```diff
@@ -409,26 +409,12 @@ export class ChordSymbol extends Modifier {
...
-   * Default text font.
-   * Choose a font family that works well with the current music engraving font.
-   * @override `Element.TEXT_FONT`.
-   */
-  static get TEXT_FONT(): Required<FontInfo> {
-    let family = 'Roboto Slab, Times, serif';
     if (Tables.currentMusicFont().getName() === 'Petaluma') {
       // Fixes Issue #1180
-      family = 'PetalumaScript, Arial, sans-serif';
+      this.textFont.family = 'PetalumaScript, Arial, sans-serif';
```
ronyeh, line 414: Should this alternate font family be stored in CommonMetrics as well? We might as well put them all in the same place instead of having `PetalumaScript, Arial, sans-serif` in a separate file.

ronyeh (changes requested):
Thanks for your hard work, Rodrigo.

I've finished reviewing this PR and have left suggestions / comments below.

Let me know what you think! I think we are moving in a good direction.

My two biggest concerns:
- `lookupMetric(...)` is not easy to understand. Also, if we structure CommonMetrics slightly differently, perhaps we can get all the font metrics in one line?
- `measureText()` requires a canvas context. Are there cases where VexFlow could be used where this is not possible? Are there environments where canvas is not available? For example, in `vexflow/demos/node/svg.js`, we just use JSDOM to output a SVG string. With the VexFlow 5 approach, these old methods will probably be broken. I am OK with temporarily breaking them, but it would be nice to discover an approach where alternate uses of VexFlow will still work fine. VexFlow 5 seems to target the browser environment only. (That's fine, and is the common case of course.) It would be cool to consider how to support those other cases somehow.

src/tables.ts
```diff
@@ -586,6 +718,41 @@ export class Tables {
...
+      if (keyParts.length > 1) {
+        keyParts[keyParts.length - 2] = keyParts[keyParts.length - 1];
+      }
+      keyParts.pop();
+    }
+
+    // Return the retrieved or default value
+    return currObj ? currObj : defaultValue;
+  }
   /**
```
ronyeh, line 756: There should be an empty line after the end of `lookupMetric()` and before this comment block.

src/chordsymbol.ts
rvilarl: This is resolved in #15

src/tables.ts
rvilarl: done

src/element.ts
rvilarl: done

src/element.ts
rvilarl: done

src/element.ts
rvilarl: It is not the shifts, it is the `xPos` and `yPos`  required by `renderText'. The default implementation would draw all the elements at the same position, overlapping with each other.
If you think that it adds value, I can include with `xPos` and `yPos` set to 0. Should I?

src/element.ts
rvilarl: done

src/tables.ts
```diff
@@ -586,6 +718,41 @@ export class Tables {
     return clefs[clef];
   }
 
+  /**
+   * Use the provided key to look up a value in this font's metrics file (e.g., bravuraMetrics.ts, petalumaMetrics.ts).
```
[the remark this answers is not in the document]
rvilarl: resolved

src/tables.ts
```diff
@@ -9,6 +9,138 @@ import { RuntimeError } from './util';
...
+    fontSize: 10,
+  },
+
+  Bend: {
+    fontFamily: 'Arial, sans-serif',
+    fontSize: 10,
+  },
+
+  ChordSymbol: {
+    fontFamily: 'Roboto Slab, Times, serif',
```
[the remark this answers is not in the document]
rvilarl: This is not a problem, we do not download the font from google, we download it from your package
https://unpkg.com/vexflow-fonts@1.0.5/

src/tables.ts
```diff
@@ -9,6 +9,138 @@ import { RuntimeError } from './util';
...
+  },
+
+  FretHandFinger: {
+    fontFamily: 'Arial, sans-serif',
+    fontSize: 9,
+    fontWeight: 'bold',
+  },
+
+  PedalMarking: {
+    fontFamily: 'Times New Roman, serif',
```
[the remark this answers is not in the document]
rvilarl: Consideration added to #14

src/element.ts
rvilarl: done

src/element.ts
rvilarl: done

src/element.ts
rvilarl: done, height also modified to be a member variable.

src/element.ts
rvilarl: I believe that we are using it on the testing activites, are not we?

src/element.ts
rvilarl: Comment corrected. Do you know of any concrete example where this is not a solution? Why do you say that "this approach (requiring a HTMLCanvasElement for measurements) is not a solution for everyone".

src/flow.ts
rvilarl: This is right. Restored.

src/staverepetition.ts
rvilarl: yes

src/stavetie.ts
rvilarl: deleted

src/tickable.ts
```diff
@@ -125,11 +119,6 @@ export abstract class Tickable extends Element {
     return this;
   }
 
-  /** Set width of note. Used by the formatter for positioning. */
-  setWidth(width: number): void {
```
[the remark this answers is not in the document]
rvilarl: Those member functions are now in `Element` and `Tickable`derives from `Element`.

src/textnote.ts
```diff
@@ -35,13 +35,6 @@ export class TextNote extends Note {
     return Category.TextNote;
   }
 
-  static TEXT_FONT: Required<FontInfo> = {
```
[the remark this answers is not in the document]
rvilarl: yes

tests/font_tests.ts
```diff
@@ -135,10 +130,7 @@ function setTextFontToGeorgia(options: TestOptions): void {
     factory.StaveNote({ keys: ['c/4', 'f/4', 'a/4'], stemDirection: -1, duration: 'q' }),
   ]);
 
-  const defaultFont = TextNote.TEXT_FONT;
-
-  // Set the default before we instantiate TextNote objects with the factory.
-  TextNote.TEXT_FONT = {
+  const defaultFont = {
```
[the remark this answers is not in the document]
rvilarl: Changed to `georgiaFont`. The way to change defaults now would be to modify CommonMetrics.

tests/font_tests.ts
```diff
@@ -148,16 +140,14 @@ function setTextFontToGeorgia(options: TestOptions): void {
...
+      .setJustification(TextNote.Justification.LEFT).setFont(defaultFont),
   ]);
 
   const formatter = factory.Formatter();
   formatter.joinVoices([voice1, voice2]).formatToStave([voice1, voice2], stave);
 
   factory.draw();
 
-  // Restore the previous default, or else it will affect the rest of the tests.
-  TextNote.TEXT_FONT = defaultFont;
```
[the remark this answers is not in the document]
rvilarl: The way to change defaults now would be to modify CommonMetrics.

src/element.ts
ronyeh: That's why I was asking about removing the `abstract` keyword. If you remove it, people can use the Element class directly, but it'll just throw errors. In this case, the error message is not super useful since it isn't telling us that xPos and yPos need to be set to a good position.

If you want Element to not be abstract, then I think draw() should have a default implementation. I'm OK with xPos and yPos being 0, and having the text element just drawn at 0,0 with default scale. Maybe that's useful for quick debugging. You could always console.warn() if you think drawing an element at 0,0 would be confusing to a developer. The warning could let the developer know that they should override the draw() method to something more useful.

src/strokes.ts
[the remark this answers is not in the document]
rvilarl: Changed to `Tables.lookupMetricFontInfo(`Strokes.text`);`

I did not follow the approach (I considered it), because I wanted to resolve the values (defaults defined a level above) at size, family... level not at font level.

With your proposal:
```js
{
 font: {
    family: 'Times New Roman, serif',
    size: 12,
    weight: 'bold',
    style: 'italic',
 },

 FretHandFinger: {
  font: {
    family: 'Arial, sans-serif',
    size: 9,
    weight: 'bold', 
   },
},
```
`FretHandFinger` would not be `italic`.

src/element.ts
rvilarl: done

src/element.ts
rvilarl: unyfied

rvilarl:
@ronyeh please look at the changes.

src/element.ts
ronyeh: @rvilarl I'm going to merge this PR but we should continue discussing the possibility of having a default draw() method that does something (instead of throwing a RuntimeError).

src/element.ts
```diff
@@ -453,27 +438,22 @@ export abstract class Element {
...
 
-  /**
-   * @returns a CSS font-style string (e.g., 'italic').
-   */
+  /** @returns a CSS font-style string (e.g., 'italic'). */
   get fontStyle(): string {
     return this.fontInfo.style;
   }
 
+  //** Set the font style. */
```
ronyeh, line 456: @rvilarl I'm going to merge this PR ... but in your next PR can you fix this minor typo. The comment starts with two slashes but should only have one slash.

src/tables.ts
```diff
@@ -586,6 +718,56 @@ export class Tables {
     return clefs[clef];
   }
 
+  /** Use the provided key to look up a FontInfo in CommonMetrics. **/
+  static lookupMetricFontInfo(key: string): Required<FontInfo> {
+    return {
+      family: Tables.lookupMetric(`${key}.fontFamily`),
+      size: Tables.lookupMetric(`${key}.fontSize`),
+      weight: Tables.lookupMetric(`${key}.fontWeight`),
+      style: Tables.lookupMetric(`${key}.fontStyle`),
```
ronyeh, line 727: @rvilarl I'm going to merge this PR
... but in your next PR consider doing `key + '.fontFamily'`. It's possible that it minimizes better. :-) You can check the output in our build/ dir, haha.

src/tables.ts
```diff
@@ -586,6 +718,56 @@ export class Tables {
...
+   * @param key is a string separated by periods (e.g., Stroke.text.fontFamily).
+   *        The following keys will be tried before using the default value:
+   *          - Stroke.text.fontFamily
+   *          - Stroke.fontFamily
+   *          - fontFamily
+   * @param defaultValue is returned if the lookup fails.
+   * @returns the retrieved value (or `defaultValue` if the lookup fails).
+   */
+  // eslint-disable-next-line
+  static lookupMetric(key: string, defaultValue?: any | number): any {
```
ronyeh, line 742: @rvilarl In the next PR, let's continue thinking about how lookupMetric() could be improved. I was particularly confused about the keyParts manipulation / pop near the end of this method.

ronyeh:
Thanks!

PR #19, opened 2023-07-11
glyph inspector using fonts directly

rvilarl:
This inspector helps a lot to see the difference between glyphs as the position of them stay the same when selecting another font.

Just a tool update, no visual differences.

### DIFF
--- a/tools/fonts/inspector/glyphs.html
+++ b/tools/fonts/inspector/glyphs.html
@@ -133,6 +133,11 @@
     <script>
       let ctx;
 
+      function codepoint2decimal(codepoint) {
+        // Return the decimal value of 'XXXX' of codepoint, which has a string format 'U+XXXX'
+        return parseInt(codepoint.substring(2), 16)
+      }
+
       function chooseFont(fontName) {
         document.getElementById('title').innerText = 'Glyph Inspector – ' + fontName;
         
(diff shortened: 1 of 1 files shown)


rvilarl:
image

tools/fonts/inspector/glyphs.html
```diff
@@ -96,54 +159,65 @@ <h1 id="title">Glyph Inspector</h1>
         ctx.setTransform(2, 0, 0, 2, 0, 0); // reset the scale to (2, 2)
 
         var x = 0;
-        var y = 50;
-        for (let glyphName in glyphs) {
+        var y = 250;
+
+        const arr = Object.keys(SMuFLGlyphInfo).sort(function(a,b) { 
+          return parseInt(SMuFLGlyphInfo[a].codepoint.substring(2), 16) - parseInt(SMuFLGlyphInfo[b].codepoint.substring(2), 16); 
+        } );
```
ronyeh, line 166: Can you add a good comment here explaining this string manipulation?

I know what it is doing since we need to translate `"codepoint": "U+E30C",` to something you can display. However, if I read this code a year from now, I'll be confused. :-)

tools/fonts/inspector/glyphs.html
```diff
@@ -96,54 +159,65 @@ <h1 id="title">Glyph Inspector</h1>
...
+        arr.forEach((glyphName, index) => {
           // Show the glyph name.
+          const code = parseInt(SMuFLGlyphInfo[glyphName].codepoint.substring(2), 16);
```
ronyeh, line 169: Same here. 

For someone who doesn't know all the details, the substring(2) might be confusing.

tools/fonts/inspector/glyphs.html
```diff
@@ -96,54 +159,65 @@ <h1 id="title">Glyph Inspector</h1>
...
           ctx.fillStyle = '#222';
+          ctx.strokeStyle = '#222';
           ctx.font = '8pt Arial';
-          ctx.fillText(glyphName, x, y + 30);
-
-          const originX = x + 100;
+          if (glyphName.length > 25) {
+            ctx.fillText(glyphName.slice(0,22)+'...', x, y + 50);
```
ronyeh, line 174: Comment here might be nice. It looks like you're breaking it up into two lines?

ronyeh:
I'll merge if you can add a few comments which will help a future reader (I suggested a few candidate places).

tools/fonts/inspector/glyphs.html
rvilarl: done

tools/fonts/inspector/glyphs.html
rvilarl: Moved to a function with comment

tools/fonts/inspector/glyphs.html
rvilarl: Moved to a function with comment

ronyeh:
Thanks!

PR #21, opened 2023-07-13
fix element comments

rvilarl:
@ronyeh this was pending

### DIFF
--- a/src/element.ts
+++ b/src/element.ts
@@ -453,7 +453,7 @@ export class Element {
     return this.fontInfo.style;
   }
 
-  //** Set the font style. */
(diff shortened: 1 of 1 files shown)


src/element.ts
```diff
@@ -528,7 +528,7 @@ export class Element {
   /** Canvas used to measure text. */
   static #txtCanvas?: HTMLCanvasElement;
 
-  //** Measure the text ussing the textFont. */
+  /** Measure the text ussing the textFont. */
```
ronyeh, line 531: Typo in 'ussing' => 'using'

rvilarl: fixed

PR #22, opened 2023-07-14
vibrato (v5)

rvilarl:
I have decided to add some fun with a second topic: vibratos.

SMuFL defines a lot of them and now it is possible to use them, see https://w3c.github.io/smufl/latest/tables/multi-segment-lines.html?highlight=vibrato#multi-segment-lines-ueaa0ueb0f

As an example:

pptr-Vibrato Harsh_Vibrato Bravura svg_current

On the other side, now it is only possible to use the SMuFL glyphs. The run-time path generated vibratos are now out.

### DIFF
--- a/src/vibrato.ts
+++ b/src/vibrato.ts
@@ -9,11 +9,8 @@ import { Tables } from './tables';
 import { Category } from './typeguard';
 
 export interface VibratoRenderOptions {
-  vibratoWidth: number;
-  waveWidth: number;
-  waveHeight: number;
-  waveGirth: number;
-  harsh: boolean;
+  code: number;
+  width: number;
 }
 
 /** `Vibrato` implements diverse vibratos. */
@@ -22,7 +19,7 @@ export class Vibrato extends Modifier {
     return Category.Vibrato;
   }
 
-  public renderOptions: VibratoRenderOptions;
+  protected renderOptions: VibratoRenderOptions;
 
   /** Arrange vibratos inside a `ModifierContext`. */
   static format(vibratos: Vibrato[], state: ModifierContextState, context: ModifierContext): boolean {
@@ -61,27 +58,32 @@ export class Vibrato extends Modifier {
 
     this.position = Modifier.Position.RIGHT;
     this.renderOptions = {
-      harsh: false,
-      vibratoWidth: 20,
-      waveHeight: 6,
-      waveWidth: 4,
-      waveGirth: 2,
+      code: 0xeab0,
+      width: 20,
     };
 
-    this.setVibratoWidth(this.renderOptions.vibratoWidth);
-  }
-
-  /** Set harsh vibrato. */
-  setHarsh(harsh: boolean): this {
-    this.renderOptions.harsh = harsh;
-    return this;
+    this.setVibratoWidth(this.renderOptions.width);
   }
 
   /** Set vibrato width in pixels. */
   setVibratoWidth(width: number): this {
-    this.renderOptions.vibratoWidth = width;
-    this.setWidth(width);
+    this.renderOptions.width = width;
+    this.text = String.fromCodePoint(this.renderOptions.code);
+    this.measureText();
+
+    const items = Math.round(this.renderOptions.width / this.getWidth());
+    for (let i = 1; i < items; i++) {
+      this.text += String.fromCodePoint(this.renderOptions.code);
+    }
+    this.measureText();
+
     return this;
+  }
+
+  /** Set vibrato code. */
+  setVibratoCode(code: number): this {
+    this.renderOptions.code = code;
+    return this.setVibratoWidth(this.renderOptions.width);
   }
 
   /** Draw the vibrato on the rendering context. */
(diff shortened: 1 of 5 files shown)


src/vibrato.ts
```diff
@@ -61,29 +58,34 @@ export class Vibrato extends Modifier {
...
-    this.setWidth(width);
+    this.renderOptions.width = width;
+    this.text = String.fromCodePoint(this.renderOptions.code);
+    this.measureText();
+
+    const items = Math.round(this.renderOptions.width / this.getWidth());
+    for (let i = 1; i < items; i++) {
+      this.text += String.fromCodePoint(this.renderOptions.code);
+    }
+    this.measureText();
```
ronyeh, line 78: It took me a while to understand this.... since the constructor() calls this function with this.renderOptions.width and then this function sets this.renderOptions.width to itself... :-)

My question: 

Is the most important part the first `measureText()`, because you have to figure out how wide the character is before you can calculate how many glyphs you need to render?

So something like this wouldn't work?
```ts
const vibratoGlyph = String.fromCodePoint(this.renderOptions.code);
// Render one or more vibrato glyphs.
let t = vibratoGlyph;
const numItems = Math.round(this.renderOptions.width / this.getWidth());
for (let i = 1; i < numItems; i++) {
  t += vibratoGlyph;
}
this.text = t;
this.measureText();
```

Do we have to call measureText() an extra time like this?

```ts
const vibratoGlyph = String.fromCodePoint(this.renderOptions.code);
// Render one or more vibrato glyphs.
let t = vibratoGlyph;

// We need to measureText() so we know the width of a single vibrato glyph.
this.text = t;
this.measureText();

const numItems = Math.round(this.renderOptions.width / this.getWidth());
for (let i = 1; i < numItems; i++) {
  t += vibratoGlyph;
}
this.text = t;
this.measureText();
```

ronyeh:
Would you mind adding a few more tests, to highlight the different possibilities possible with this approach?

It looks like there are a lot of options here:
https://w3c.github.io/smufl/latest/tables/multi-segment-lines.html?highlight=vibrato#multi-segment-lines-ueaa0ueb0f

It would be fun to see a couple more of them (set the vibrato wide enough so we see at least 3-glyphs in a row, to visualize the multi segment lines).

src/vibrato.ts
```diff
@@ -9,11 +9,8 @@ import { Tables } from './tables';
...
-  vibratoWidth: number;
-  waveWidth: number;
-  waveHeight: number;
-  waveGirth: number;
-  harsh: boolean;
+  code: number;
+  width: number;
 }
 
 /** `Vibrato` implements diverse vibratos. */
```
ronyeh, line 16: Can we expand this comment?

I think we need to point developers to: https://w3c.github.io/smufl/latest/tables/multi-segment-lines.html

src/vibrato.ts
```diff
@@ -61,29 +58,34 @@ export class Vibrato extends Modifier {
...
+    for (let i = 1; i < items; i++) {
+      this.text += String.fromCodePoint(this.renderOptions.code);
+    }
+    this.measureText();
+
     return this;
   }
 
+  /** Set vibrato code. */
+  setVibratoCode(code: number): this {
```
ronyeh, line 84: It is not clear what codes are supported. We should point developers to https://w3c.github.io/smufl/latest/tables/multi-segment-lines.html

And teach the developer if they want a:

```
U+EAB5
wiggleWavy
Wavy line segment
```

Then they need to pass in the code `0xeab5`.

tests/vibrato_tests.ts
```diff
@@ -105,7 +105,7 @@ function withBend(options: TestOptions, contextBuilder: ContextBuilder): void {
     tabNote({
       positions: [{ str: 2, fret: 10 }],
       duration: 'h',
-    }).addModifier(new Vibrato().setVibratoWidth(120).setHarsh(true), 0),
+    }).addModifier(new Vibrato().setVibratoWidth(120).setVibratoCode(0xeae2), 0),
```
ronyeh, line 108: To someone reading these tests, it's unclear where the magic codes come from.

Maybe you can use the friendly name (via a local variable `const wiggleVibratoLargeFastest = 0xeae2;`). And add the comment pointing the developer to the relevant SMuFL page for more codes. It's OK if the tests are verbose. They double as a way to teach developers how to use our API.

src/vibrato.ts
ronyeh: Maybe we need a block comment at the top of the file with a list of the recommended codes and their friendly SMuFL names. Just reading the code with these hex code points is confusing. :-)

rvilarl:
@ronyeh thanks a lot for the comments, I will act on them soon but I am travelling this week. 

@ronyeh @mscuthbert @AaronDavidNewman By the way in the SMuFL page at the end there are vibrates that combine several glyphs. We could also support them, what do you think?

rvilarl:
@ronyeh thanks a lot for the comments, I will act on them soon but I am travelling this week. 

@ronyeh @mscuthbert @AaronDavidNewman By the way in the SMuFL page at the end there are vibrates that combine several glyphs. We could also support them, what do you think?

ronyeh:
Take your time! In fact, the rest of us need the extra time to catch up, since you've been investigating this for a while and much of this is new to us. So most times you will actually be waiting on us (to read your code and provide feedback).

Regarding mixing & matching vibratos... let's skip that for now and keep it simple. We don't need to support every edge case until someone requests it.

src/vibrato.ts
rvilarl: The first call to `measureText` calculates the `width` of one glyph. This is used to evaluate how many glyphs we need. The second call evaluates the actual width with `numImtems` Glyphs.

ronyeh:
Looks good. I might offer some of my previous feedback on this as a PR.

ronyeh:
I realize I'm probably merging the PRs out of order.... :-)

Vibrato probably depends on the SMuFL fonts being properly loaded by the previous PR.

It's OK, we'll have #15 reviewed and merged soon, I think.

PR #25, opened 2023-07-20
articulation (v5)

rvilarl:
Articulations using font. Minor visual differences.

### DIFF
--- a/src/articulation.ts
+++ b/src/articulation.ts
@@ -3,7 +3,6 @@
 // MIT License
 
 import { Builder } from './easyscore';
-import { Glyph } from './glyph';
 import { Modifier } from './modifier';
 import { ModifierContextState } from './modifiercontext';
 import { Note } from './note';
@@ -12,12 +11,12 @@ import { Stem } from './stem';
 import { StemmableNote } from './stemmablenote';
 import { Tables } from './tables';
 import { Category, isGraceNote, isStaveNote, isStemmableNote, isTabNote } from './typeguard';
-import { defined, log, RuntimeError } from './util';
+import { log, RuntimeError } from './util';
 
 export interface ArticulationStruct {
-  code?: string;
-  aboveCode?: string;
-  belowCode?: string;
+  code?: number;
+  aboveCode?: number;
+  belowCode?: number;
   betweenLines: boolean;
 }
 
@@ -175,14 +174,12 @@ export class Articulation extends Modifier {
   protected static readonly INITIAL_OFFSET: number = -0.5;
 
   /** Articulation code provided to the constructor. */
-  readonly type: string;
+  readonly type: string | number;
 
-  public renderOptions: { fontScale: number };
-  // articulation defined calling reset in constructor
-  protected articulation!: ArticulationStruct;
-  // glyph defined calling reset in constructor
-  protected glyph!: Glyph;
+  protected articulation: ArticulationStruct;
 
+  protected heightShift = 0;
+  protected height = 0;
   /**
    * FIXME:
    * Most of the complex formatting logic (ie: snapping to space) is
@@ -208,14 +205,11 @@ export class Articulation extends Modifier {
     let maxGlyphWidth = 0;
 
     const getIncrement = (articulation: Articulation, line: number, position: number) =>
-      roundToNearestHalf(
-        getRoundingFunction(line, position),
-        defined(articulation.glyph.getMetrics().height) / 10 + margin
-      );
+      roundToNearestHalf(getRoundingFunction(line, position), articulation.height / 10 + margin);
 
     articulations.forEach((articulation) => {
       const note = articulation.checkAttachedNote();
-      maxGlyphWidth = Math.max(note.getGlyphProps().getWidth(), maxGlyphWidth);
+      maxGlyphWidth = Math.max(note.getGlyphWidth(), maxGlyphWidth);
       let lines = 5;
       const stemDirection = note.hasStem() ? note.getStemDirection() : Stem.UP;
       let stemHeight = 0;
@@ -298,35 +292,35 @@ export class Articulation extends Modifier {
    * @param type entry in `Vex.Flow.articulationCodes` in `tables.ts` or Glyph code.
    *
    * Notes (by default):
-   * - Glyph codes ending with 'Above' will be positioned ABOVE
-   * - Glyph codes ending with 'Below' will be positioned BELOW
+   * - Even codes will be positioned ABOVE
+   * - Odd codes will be positioned BELOW
    */
-  constructor(type: string) {
+  constructor(type: string | number) {
     super();
 
     this.type = type;
     this.position = ABOVE;
-    this.renderOptions = {
-      fontScale: Tables.NOTATION_FONT_SCALE,
-    };
+    if (typeof this.type == 'number') {
+      if (this.type % 2 == 0) this.position = ABOVE;
+      else this.position = BELOW;
+    }
 
+    this.articulation = { betweenLines: false };
     this.reset();
   }
 
   protected reset(): void {
-    this.articulation = Tables.articulationCodes(this.type);
+    if (typeof this.type == 'string') this.articulation = Tables.articulationCodes(this.type);
     // Use type as glyph code, if not defined as articulation code
-    if (!this.articulation) {
+    if (typeof this.type == 'number') {
       this.articulation = { code: this.type, betweenLines: false };
-      if (this.type.endsWith('Above')) this.position = ABOVE;
-      if (this.type.endsWith('Below')) this.position = BELOW;
     }
     const code =
-      (this.position === ABOVE ? this.articulation.aboveCode : this.articulation.belowCode) || this.articulation.code;
-    this.glyph = new Glyph(code ?? '', this.renderOptions.fontScale);
-    defined(this.glyph, 'ArgumentError', `Articulation not found: ${this.type}`);
-
-    this.setWidth(defined(this.glyph.getMetrics().width));
+      (this.position === ABOVE ? this.articulation.aboveCode : this.articulation.belowCode) ||
+      this.articulation.code ||
+      0x0000;
+    this.text = String.fromCharCode(code);
+    this.measureText();
   }
 
   /** Set if articulation should be rendered between lines. */
@@ -342,7 +336,7 @@ export class Articulation extends Modifier {
     this.setRendered();
 
     const index = this.checkIndex();
-    const { position, glyph, textLine } = this;
+    const { position, textLine } = this;
     const canSitBetweenLines = this.articulation.betweenLines;
 
     const stave = note.checkStave();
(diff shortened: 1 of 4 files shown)


ronyeh:
For this and future VexFlow 5 PRs, can you attach a sample image of the output? It will be helpful for establishing context for the PR (especially if someone other than me, you, Aaron decide to drop in).

src/tables.ts
```diff
@@ -598,49 +598,53 @@ const accidentalColumns: Record<number, { [name: string]: number[] }> = {
 };
 
 const articulations: Record<string, ArticulationStruct> = {
-  'a.': { code: 'augmentationDot', betweenLines: true }, // Staccato
+  'a.': { code: 0xe1e7 /*augmentationDot*/, betweenLines: true }, // Staccato
   av: {
-    aboveCode: 'articStaccatissimoAbove',
-    belowCode: 'articStaccatissimoBelow',
+    aboveCode: 0xe4a6 /*articStaccatissimoAbove*/,
```
ronyeh, line 603: Thanks. These comments are very helpful! :-)

src/tables.ts
```diff
@@ -598,49 +598,53 @@ const accidentalColumns: Record<number, { [name: string]: number[] }> = {
...
+  'a@': { aboveCode: 0xe4c0 /*fermataAbove*/, belowCode: 0xe4c1 /*fermataBelow*/, betweenLines: false }, // Fermata
+  'a@a': { code: 0xe4c0 /*fermataAbove*/, betweenLines: false }, // Fermata above staff
+  'a@u': { code: 0xe4c1 /*fermataBelow*/, betweenLines: false }, // Fermata below staff
+  'a@s': { aboveCode: 0xe4c4 /*fermataShortAbove*/, belowCode: 0xe4c5 /*fermataShortBelow*/, betweenLines: false }, // Fermata short
+  'a@as': { code: 0xe4c4 /*fermataShortAbove*/, betweenLines: false }, // Fermata short above staff
+  'a@us': { code: 0xe4c5 /*fermataShortBelow*/, betweenLines: false }, // Fermata short below staff
+  'a@l': { aboveCode: 0xe4c6 /*fermataLongAbove*/, belowCode: 0xe4c7 /*fermataLongBelow*/, betweenLines: false }, // Fermata long
+  'a@al': { code: 0xe4c6 /*fermataLongAbove*/, betweenLines: false }, // Fermata long above staff
+  'a@ul': { code: 0xe4c7 /*fermataLongBelow*/, betweenLines: false }, // Fermata long below staff
+  'a@vl': {
```
ronyeh, line 638: I admit I am not very familiar with this area of the codebase. 

What are these special codes like a@ul and a@uvl ?

Are they VexFlow specific shorthand for specifying Articulations? Since we are breaking backwards compatibility with V5, perhaps we can do a separate PR to improve this. The magic shorthand isn't easy to understand.

rvilarl: yes, they are vexflow specific shorthands.

rvilarl: Probably it would be better to use the code directly and calculate betweenLines on run time based on the glyph height.

rvilarl:
Several examples, minor differences (current first)
pptr-Articulation Articulation___Accent_Tenuto Bravura svg_current
pptr-Articulation Articulation___Accent_Tenuto Bravura svg_reference
pptr-Articulation Articulation___Fermata_Above_Below Bravura svg_current
pptr-Articulation Articulation___Fermata_Above_Below Bravura svg_reference
pptr-Articulation Articulation___Fermata_Long_Above_Below Bravura svg_current
pptr-Articulation Articulation___Fermata_Long_Above_Below Bravura svg_reference
pptr-Articulation Articulation___Fermata_Short_Above_Below Bravura svg_current
pptr-Articulation Articulation___Fermata_Short_Above_Below Bravura svg_reference
pptr-Articulation Articulation___Fermata_Very_Long_Above_Below Bravura svg_current
pptr-Articulation Articulation___Fermata_Very_Long_Above_Below Bravura svg_reference
pptr-Articulation Articulation___Inline_Multiple Bravura svg_current
pptr-Articulation Articulation___Inline_Multiple Bravura svg_reference
pptr-Articulation Articulation___Marcato_L_H__Pizzicato Bravura svg_current
pptr-Articulation Articulation___Marcato_L_H__Pizzicato Bravura svg_reference
pptr-Articulation Articulation___Snap_Pizzicato_Fermata Bravura svg_current
pptr-Articulation Articulation___Snap_Pizzicato_Fermata Bravura svg_reference
pptr-Articulation Articulation___Staccato_Staccatissimo Bravura svg_current
pptr-Articulation Articulation___Staccato_Staccatissimo Bravura svg_reference
pptr-Articulation Articulation___Up_stroke_Down_Stroke Bravura svg_current
pptr-Articulation Articulation___Up_stroke_Down_Stroke Bravura svg_reference
pptr-Articulation Articulation___Vertical_Placement Bravura svg_current
pptr-Articulation Articulation___Vertical_Placement Bravura svg_reference
pptr-Articulation Articulation___Vertical_Placement__Glyph_codes_ Bravura svg_current
pptr-Articulation Articulation___Vertical_Placement__Glyph_codes_ Bravura svg_reference

tests/articulation_tests.ts
```diff
@@ -250,53 +250,59 @@ function verticalPlacement2(options: TestOptions, contextBuilder: ContextBuilder
...
+      )
+      .addModifier(new Articulation(0xe4a5 /*articTenutoBelow*/).setBetweenLines(), 0)
+      .addModifier(new Articulation(0xe4c1 /*fermataBelow*/), 0),
     staveNote({ keys: ['a/5'], duration: 'q', stemDirection: Stem.DOWN })
-      .addModifier(new Articulation('fermataAbove'), 0)
-      .addModifier(new Articulation('augmentationDot').setPosition(ModifierPosition.ABOVE), 0)
-      .addModifier(new Articulation('articTenutoAbove'), 0),
+      .addModifier(new Articulation(0xe4c0 /*fermataAbove*/), 0)
+      .addModifier(new Articulation(0xe1e7 /*augmentationDot*/).setPosition(ModifierPosition.ABOVE), 0)
+      .addModifier(new Articulation(0xe4a4 /*articTenutoAbove*/), 0),
```
ronyeh, line 282: Is there a benefit to using numeric codepoints? Later on we have to translate it via String.fromCodePoint() or String.fromCharCode().

Should we refactor this class to accept a string, and then require the developer to pass in the correct SMuFL glyph directly?

```
String.fromCharCode(0xE4C0) === '\uE4C0'
```

So instead of using numbers, we could use the `\uXXXX` escape code for unicode characters.

rvilarl: done

src/articulation.ts
```diff
@@ -298,18 +292,20 @@ export class Articulation extends Modifier {
    * @param type entry in `Vex.Flow.articulationCodes` in `tables.ts` or Glyph code.
    *
    * Notes (by default):
-   * - Glyph codes ending with 'Above' will be positioned ABOVE
-   * - Glyph codes ending with 'Below' will be positioned BELOW
+   * - Even codes will be positioned ABOVE
+   * - Odd codes will be positioned BELOW
```
ronyeh, line 296: Maybe we need to reference the appropriate SMuFL glyph page here? It seems arbitrary that Even is above and Odd is below, but if you look at the SMuFL code points, it makes more sense.

src/articulation.ts
```diff
@@ -298,18 +292,20 @@ export class Articulation extends Modifier {
...
     super();
 
     this.type = type;
     this.position = ABOVE;
-    this.renderOptions = {
-      fontScale: Tables.NOTATION_FONT_SCALE,
-    };
+    if (!Tables.articulationCodes(this.type)) {
+      if (this.type.charCodeAt(0) % 2 == 0) this.position = ABOVE;
+      else this.position = BELOW;
```
ronyeh, line 305: Let's make sure we do not need to use `String.codePointAt()` instead:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt

charCodeAt doesn't support the entire range of unicode code points.

(I guess passing in numeric code points would make this check a bit easier.... but I do think using the SMuFL glyph directly is probably better.)

ronyeh: We should be using codePointAt().

U+1D12B is an alternate codepoint for accidentalDoubleFlat.

0x1D12B is odd.

'\u{1D12B}'.charCodeAt(0) % 2 is 0, which implies it is even. This is incorrect.

'\u{1D12B}'.codePointAt(0) % 2 is 1, which means it is odd. This is correct.

src/articulation.ts
```diff
@@ -318,15 +314,13 @@ export class Articulation extends Modifier {
...
     }
     const code =
-      (this.position === ABOVE ? this.articulation.aboveCode : this.articulation.belowCode) || this.articulation.code;
-    this.glyph = new Glyph(code ?? '', this.renderOptions.fontScale);
-    defined(this.glyph, 'ArgumentError', `Articulation not found: ${this.type}`);
-
-    this.setWidth(defined(this.glyph.getMetrics().width));
+      (this.position === ABOVE ? this.articulation.aboveCode : this.articulation.belowCode) ||
+      this.articulation.code ||
+      '\u0000';
```
ronyeh, line 321: I don't understand this part.

Is it doing some sanity check with fallback? Why is the fallback the unicode codepoint 0? Should it be an empty string instead?

src/element.ts
```diff
@@ -2,6 +2,7 @@
 // MIT License
 // @author Mohit Cheppudira
 
+import { BEAM_BOTH } from './beam';
```
ronyeh, line 5: Is this import used? I don't see it below. Why does element base class need to know about beams?

tests/articulation_tests.ts
```diff
@@ -250,53 +250,59 @@ function verticalPlacement2(options: TestOptions, contextBuilder: ContextBuilder
 
   const notes = [
     staveNote({ keys: ['f/4'], duration: 'q' })
-      .addModifier(new Articulation('fermataBelow'), 0)
-      .addModifier(new Articulation('augmentationDot').setPosition(ModifierPosition.BELOW), 0)
-      .addModifier(new Articulation('articTenutoBelow'), 0),
+      .addModifier(new Articulation('\ue4c1' /*fermataBelow*/), 0)
+      .addModifier(new Articulation('\ue1e7' /*augmentationDot*/).setPosition(ModifierPosition.BELOW), 0)
+      .addModifier(new Articulation('\ue4a5' /*articTenutoBelow*/), 0),
```
ronyeh, line 255: I think it is OK that we are using the unicode codepoints directly, for now.

However, developers would have to look up the code points in the SMuFL spec each time. It would be nice if there was an easy (readable) way to specify the glyphs we want. Maybe we can figure out the const enum thing at some point.

src/tables.ts
ronyeh: Does anyone use these shorthand codes? I don't see it used in our tests.

We can keep it for now I suppose (for legacy reasons) but this secret shorthand isn't that readable anyways. :-)

ronyeh:
OK I'm done reviewing this. Please see my comments above.

src/tables.ts
rvilarl: They are used a lot in articulation_tests.ts

src/articulation.ts
rvilarl: done

src/element.ts
rvilarl: removed

tests/articulation_tests.ts
rvilarl: Agree.

src/articulation.ts
rvilarl: I used 0 becuase this code on some fonts displays a square or 'NOGLYPH' showing that there is a problem.

src/tables.ts
AaronDavidNewman: >> Does anyone use these shorthand codes? I don't see it used in our tests.

>> We can keep it for now I suppose (for legacy reasons) but this secret shorthand isn't that readable anyways. :-)

> They are used a lot in articulation_tests.ts

They are also used for easyScore, which is why they're so cryptic.

If anyone uses 'fermataLongBelow' because that's how it was done in the test cases, will it still work? 

What happens if someone passes in a smufl code that's not an articulation - will it still render?  I don't know that we should prevent it, just curious.

PR #27, opened 2023-07-24
stavetempo (v5)

rvilarl:
Adding a new case with lot of code saving and also visual differences.

### DIFF
--- a/src/stavetempo.ts
+++ b/src/stavetempo.ts
@@ -1,11 +1,9 @@
 // Copyright (c) 2023-present VexFlow contributors: https://github.com/vexflow/vexflow/graphs/contributors
 // @author: Radosaw Eichler 2012
 
-import { Glyph } from './glyph';
 import { Stave } from './stave';
 import { StaveModifier, StaveModifierPosition } from './stavemodifier';
 import { Tables } from './tables';
-import { TextFormatter } from './textformatter';
 import { Category } from './typeguard';
 
 export interface StaveTempoOptions {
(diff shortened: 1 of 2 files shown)


rvilarl:
current first

Bravura
pptr-Stave Tempo_Test Bravura svg_current
pptr-Stave Tempo_Test Bravura svg_reference
Gonville
pptr-Stave Tempo_Test Gonville svg_current
pptr-Stave Tempo_Test Gonville svg_reference
Leland
pptr-Stave Tempo_Test Leland svg_current
pptr-Stave Tempo_Test Leland svg_reference
Petaluma
pptr-Stave Tempo_Test Petaluma svg_current
pptr-Stave Tempo_Test Petaluma svg_reference

ronyeh:
I'm not an expert on how they "should look". I think the bigger ones look better, personally. 

Also does the current one have extra spacing inside "Andante&nbsp;&nbsp;&nbsp;grazioso"?

ronyeh:
Also, this was a problem with the old version as well, but are the whole notes supposed to be below the equal sign? It looks too low for me, but once again, I'm not the authority on how it is supposed to look.

src/stavetempo.ts
```diff
@@ -20,107 +18,52 @@ export class StaveTempo extends StaveModifier {
...
-    return this;
+    this.setXShift(10);
+    this.setYShift(shiftY);
   }
 
-  setShiftX(x: number): this {
-    this.shiftX = x;
-    return this;
-  }
+  private durationToCode: Record<string, string> = {
```
ronyeh, line 31: Should we continue to use the JS prefix for private members, i.e. `#durationToCode`?

ronyeh:
To be honest, I'm not sure I like the new look :-)

The glyphs are too small for my taste, and the augmentation dots are spaced too far apart (especially in the last Petaluma example, with multiple dots). I liked the old way better.

Also, I'm not sure the glyphs are aligned well next to the equal sign. It seems like some of the right stems are almost touching the equal sign.

Finally, the whole note seems to be way too low. But I think that's more a function of the smaller glyphs. And I think it isn't really the common case, as usually metronome marks don't specify a BPM for a whole note.

I like the massive simplification of the code due to the new codepoint based approach.

However, is there a way to make the output more pleasing to the eye (or more similar to v4, at least). I'm worried that if we stray too far from v4 visually, that the visual changes will add up and make the final scores look bad.

rvilarl:
Back to separated elements.

Bravura
pptr-Stave Tempo_Test Bravura svg_current
Gonville
pptr-Stave Tempo_Test Gonville svg_current
Leland (this font does not include `metNote32ndUp` and the glyph defaults to Bravura)
pptr-Stave Tempo_Test Leland svg_current
Petaluma
pptr-Stave Tempo_Test Petaluma svg_current

src/stavetempo.ts
rvilarl: done

ronyeh:
Can we investigate why the Leland tempo glyphs are so big? (I'm not talking about the missing glyph.)

Some of them look bigger than the regular notes on the staff.

I'm happy to merge this, but will file a couple issues to investigate:

1) Leland glyphs are possibly too big here. Leland is the default font in MuseScore. Are their metronome / tempo glyphs as big as ours?

2) I'm not sure Roboto Slab fits well as a default text font. At some point we should move to Academico or even back to Times as the default font, but allow people like Aaron to choose Roboto Slab if they wish.

rvilarl:
@ronyeh, for 2, could you release a version of vexflow-fonts?

rvilarl:
@ronyeh this is how it looks in MuseScore

image

Issue #29, opened 2023-07-24
String.fromCharCode(...) doesn't support code points above 0xFFFF


ronyeh:
If we use any SMuFL code points above that, we need to use: `String.fromCodePoint()`

There are 219 SMuFL glyphs that won't work with `String.fromCharCode()`:

```
accidentalDoubleFlatAlternate = 0x1d12b,
accidentalDoubleSharpAlternate = 0x1d12a,
accidentalQuarterToneFlat4Alternate = 0x1d133,
accidentalQuarterToneFlatArrowUpAlternate = 0x1d12c,
accidentalQuarterToneFlatNaturalArrowDownAlternate = 0x1d12f,
accidentalQuarterToneSharp4Alternate = 0x1d132,
accidentalQuarterToneSharpArrowDownAlternate = 0x1d131,
accidentalQuarterToneSharpNaturalArrowUpAlternate = 0x1d12e,
accidentalThreeQuarterTonesFlatArrowDownAlternate = 0x1d12d,
accidentalThreeQuarterTonesSharpArrowUpAlternate = 0x1d130,
analyticsEndStimmeAlternate = 0x1d1a8,
analyticsHauptstimmeAlternate = 0x1d1a6,
analyticsNebenstimmeAlternate = 0x1d1a7,
arpeggiatoDownAlternate = 0x1d184,
arpeggiatoUpAlternate = 0x1d183,
articAccentAboveAlternate = 0x1d17b,
articAccentStaccatoAboveAlternate = 0x1d181,
articMarcatoAboveAlternate = 0x1d17f,
articMarcatoStaccatoAboveAlternate = 0x1d180,
articStaccatissimoAboveAlternate = 0x1d17e,
articStaccatoAboveAlternate = 0x1d17c,
articTenutoAboveAlternate = 0x1d17d,
articTenutoStaccatoAboveAlternate = 0x1d182,
augmentationDotAlternate = 0x1d16d,
barlineDashedAlternate = 0x1d104,
barlineDoubleAlternate = 0x1d101,
barlineFinalAlternate = 0x1d102,
barlineReverseFinalAlternate = 0x1d103,
barlineShortAlternate = 0x1d105,
barlineSingleAlternate = 0x1d100,
braceAlternate = 0x1d114,
bracketAlternate = 0x1d115,
brassBendAlternate = 0x1d189,
brassDoitShortAlternate = 0x1d185,
brassFallLipShortAlternate = 0x1d186,
[...]

PR #30, opened 2023-07-24
Use const enum to have readable glyph names that are inlined as unicode code points and unicode strings during compilation.

ronyeh:
This is a work in progress as I have discovered some issues with our use of const enums, and how it may conflict with the way we compile our library.

Anyways.... at least in the ES modules output, this works perfectly fine.

This PR adds two const enums, `SMuFLGlyph` and `SMuFLCodePoint`, which contain all the smufl glyphs and hex code points.

The big idea is that in our code, we can use readable names like: `SMuFLGlyph.csymMinor` 

The compiled output should ideally make no reference to these enums. All uses of the enums should be erased and replaced by their values (similar in spirit to C preprocessor `#define` constants).

So `SMuFLGlyph.csymMinor` => `"\uE874"`.

I have verified that this works great in the ESM build of our library. This means we do not need to use `String.fromCharCode()`, which allows us to avoid the issues noted in #29 .

By using these const enums, we will be using the unicode characters directly. No extra step in translating the code points to characters.

If for some reason we really do need the code point, this PR includes a const enum of all the code points (`SMuFLCodePoint`).

I'd like to hear any and all comments... both for an against this approach. :-)


ronyeh:
The biggest drawback right now is that our webpack exports do not erase the const enums... and the enums are treated as regular enums.

We have used const enums in the past.... so it's possible that we haven't been getting the "full performance benefit" of those const enums.

It is possible that our efforts to speed up our build (by setting `transpileOnly: true`) is breaking the const enums. See: https://github.com/microsoft/TypeScript/issues/16671#issuecomment-344429165 


Regardless, it would be super cool if we could figure this out. 

Maybe the right answer is to emit ES modules first, and then run the webpack build process on the ESM files? That way, we are "compiling" the ESM files, which already have the const enums erased.

ronyeh:
Even if we don't go with the const enum approach in the end, I believe we can avoid converting code points to strings in most cases.

We should just use unicode string literals like: `'\uE874'`.

ronyeh:
Remember that the `\uXXXX` escape code only works for code points less than or equal to 0xFFFF. For bigger code points, we need to use this form:

```
'\u{1d12b}'
// accidentalDoubleFlat alternate code point is > 0xFFFF
```

src/typeguard.ts
```diff
@@ -8,7 +8,7 @@ import { Dot } from './dot';
 import { GraceNote } from './gracenote';
 import { GraceNoteGroup } from './gracenotegroup';
 import { Note } from './note';
-import { RenderContext } from './rendercontext.js';
+import { RenderContext } from './rendercontext';
```
ronyeh, line 11: This fixes an error in the import. It should not include the .js extension.

src/chordsymbol.ts
```diff
@@ -360,7 +361,7 @@ export class ChordSymbol extends Modifier {
       symbolModifier: SymbolModifiers;
     }> = {}
   ): this {
-    return this.addText(String.fromCharCode(0xe874 /*csymMinor*/, 0xe874 /*csymMinor*/), params);
+    return this.addText(SMuFLGlyph.csymMinor + SMuFLGlyph.csymMinor, params);
```
ronyeh, line 364: In the ESM build, this line is successfully changed to:

```
return this.addText("\uE874" + "\uE874", params);
```

src/vibrato.ts
```diff
@@ -4,7 +4,6 @@
 import { Bend } from './bend';
 import { Modifier } from './modifier';
 import { ModifierContext, ModifierContextState } from './modifiercontext';
-import { RenderContext } from './rendercontext';
```
ronyeh, line 7: Unused import.

tests/vibrato_tests.ts
```diff
@@ -67,11 +69,11 @@ function harsh(options: TestOptions, contextBuilder: ContextBuilder): void {
         { str: 4, fret: 9 },
       ],
       duration: 'h',
-    }).addModifier(new Vibrato().setVibratoCode(0xeae2), 0),
+    }).addModifier(new Vibrato().setVibratoCode(SMuFLCodePoint.wiggleVibratoLargeFastest /* 0xeae2 */), 0),
```
ronyeh, line 72: @rvilarl Should we be relying on code points? Why not just accept the unicode character directly?

In this case, we would use: `'\ueae2'`

rvilarl:
Hi Ron,
I agree that we should move to the scape codes. The enums are no doubt easier to understand but we should make sure that the size of the library does not grow too much.  The new file is very big.

ronyeh:
I will make sure this approach does not add any size to the library. The
idea behind const enum is that it is supposed to be erased by TypeScript
during compilation, and replaced by the constant escape code. Anything that
is unused should be completely erased.

However, I inspected the webpack output and there is some evidence of some
parts of the enum remaining. The ESM output is perfect, as I expected.

Needs more investigation. In the meantime, I will suggest using the \u and
\u{} escape codes directly, and avoiding String.fromCharCode if possible.
We can just put the comment with the smufl glyph name next to each escape
code, as you have been doing. That's probably good enough for now.

ronyeh:
https://github.com/evanw/esbuild/blob/main/CHANGELOG-2021.md#0149


Maybe we will have to switch from webpack to rebuild to make it work.

There might be a webpage plugin, but I am also a fan of esbuild for other reasons so maybe it is worth the trouble.

ronyeh:
Closing this one for now, but I have been investigating `esbuild` in another branch and it is very promising.

1) It's fast.
2) It seems to handle const enums in the way we expect.

Issue #23, opened 2023-07-20
Use friendly names for SMUFL glyphs (e.g., use a const enum for codepoints)


ronyeh:
I'll assign this to myself.

Consider using `const enum` to allow us to reference SMuFL code points in a more readable way.

See: https://www.typescriptlang.org/docs/handbook/enums.html#const-enums
> ...unlike regular enums they are completely removed during compilation. Const enum members are inlined at use sites. 

See: https://github.com/w3c/smufl/blob/gh-pages/metadata/glyphnames.json

Using the `const enum` could look like:

```
smufl.coda => E048
smufl.codaSquare => 0xE049
smufl.codaAlternate => 0x1D10C
```

I wasn't sure how to capitalize the acronym.

The official format is `SMuFL` with the lowercase `u`. We could simplify it with all caps: `SMUFL` or only do the first letter... `Smufl`. Or all lowercase?

ronyeh:
See this discussion: https://github.com/vexflow/vexflow/pull/48/files#r1298762764

ronyeh:
Linking to the discussion here: https://github.com/vexflow/vexflow/pull/35#discussion_r1313818613

ronyeh:
I'm also going to link this exploratory work here: https://github.com/vexflow/vexflow/pull/30

ronyeh:
@rvilarl and @0xfe and anyone else who might be interested, such as @AaronDavidNewman 

We are discussing how we can best use friendly readable SMuFL names in the VexFlow 5 library, without hugely inflating the library size if possible.

Mohit left some comments here: https://github.com/vexflow/vexflow/pull/35#discussion_r1314252447

I see a few options:

1. Include the entire SMuFL friendly names => code points mapping in our library. In the grand scheme of things it probably won't slow down websites all that much. Most web apps are huge these days.
2. Include only the subset of SMuFL codes that we use internally in VexFlow. This is much much smaller, because the entire SMuFL spec includes really rare glyphs that everyday scores won't use.
3. Don't include any SMuFL friendly names. We only accept Unicode characters. This results in the lightest library which some folks like Simon like. We could offload the friendly names into an external library.

Also, we need to continue exploration of `const enum` in TypeScript. The basic idea is that they act like C `#define` where TypeScript compilers will remove the variable name during compilation, and replace the long friendly names with the unicode characters directly.

More to come....

AaronDavidNewman:
[...]

PR #35, opened 2023-07-30
accidental (v5)

rvilarl:
This PR bases on #34 . Once #34 is merged I will rebase and the number of changes will be reduced.


rvilarl:
Some examples (current first)
Bravura
pptr-Accidental Accidental_Arrangement_Special_Cases Bravura svg_current
pptr-Accidental Accidental_Arrangement_Special_Cases Bravura svg_reference
Gonville
pptr-Accidental Accidental_Arrangement_Special_Cases Gonville svg_current
pptr-Accidental Accidental_Arrangement_Special_Cases Gonville svg_reference
Leland
pptr-Accidental Accidental_Arrangement_Special_Cases Leland svg_current
pptr-Accidental Accidental_Arrangement_Special_Cases Leland svg_reference
Petaluma
pptr-Accidental Accidental_Arrangement_Special_Cases Petaluma svg_current
pptr-Accidental Accidental_Arrangement_Special_Cases Petaluma svg_reference

tests/accidental_tests.ts
```diff
@@ -286,6 +281,268 @@ function basic(options: TestOptions): void {
...
+  '\uE406',
+  '\uE407',
+  '\uE408',
+  '\uE409',
+  '\uE40A',
+  '\uE40B',
+  '\uE284',
+  '\uE285',
+  '\uE47C',
+];
```
0xfe, line 544: Probably cleaner to write this like so:

```typescript
function genAccidentals(): string[] {
    let accs = ['#', '##', 'b', /* ... */];
    for (let u = 0xE300; u < 0xE40B; u++) {
        accs.push(String.fromCharCode(u));
    }
    accs.push('\uE284', '\uE285', /* ... */);
    return accs
}

const accidentals: string[] = genAccidentals();
```

0xfe (commented):
Some minor comments, since I've been away from this codebase for a while.

tests/accidental_tests.ts
```diff
@@ -643,39 +888,43 @@ function sagittal(options: TestOptions): void {
...
+      .StaveNote({ keys: ['c/4', 'e/4', 'g/4', 'c/5'], duration: '8' })
+      .addModifier(accid('\uE303' /*accSagittal5CommaDown*/), 1),
 
     f
       .StaveNote({ keys: ['c/4', 'e/4', 'g/4', 'b/4'], duration: '8' })
       .addModifier(accid('b'), 1)
-      .addModifier(accid('accSagittal7CommaDown'), 1)
-      .addModifier(accid('accSagittal11LargeDiesisDown'), 3),
+      .addModifier(accid('\uE305' /*accSagittal7CommaDown*/), 1)
+      .addModifier(accid('\uE30D' /*accSagittal11LargeDiesisDown*/), 3),
```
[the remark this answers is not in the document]
rvilarl: In Vexflow 5 it is now possible to use all the SMuFL codes. The dictionary would be very big. @ronyeh is conisdering what to do in #23

tests/accidental_tests.ts
rvilarl: This a good idea! I will use it to go through all the accidentals in the SMuFL spec.

rvilarl: done

rvilarl:
Complete set of accidentals now displayed after @0xfe review comment
Accidental Cautionary_Accidental Bravura svg_current

ronyeh:
@rvilarl FYI this is the holiday weekend, so I might not get to more reviews until middle of next week.

rvilarl:
> @rvilarl FYI this is a holiday weekend, so I might not get to more reviews until middle of next week.

Have a good weekend! After merging this one we should investigate the performance.

tests/accidental_tests.ts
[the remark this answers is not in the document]
ronyeh: Ideally, our APIs could accept both character codes and SMuFL glyph names.

Once we are done with converting the v4 codebase to the v5 full SMuFL approach, we can investigate how to do this (or if it's better altogether to just use the friendly SMuFL glyph names).

Many of our APIs also support the classic VexFlow shorthands (e.g., ## instead of accidentalDoubleSharp).

Would be cool if we supported the classic shorthands + friendly SMuFL glyph names + unicode characters (which is the most direct method).

But we'll have to see how the library size is impacted if we support all the SMuFL names.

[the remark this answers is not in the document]
0xfe: +1 to accepting all three types. Ideally, in the VexFlow codebase, we end up using the readable names as much as possible.

The following would be nice:

```typescript
note.addModifier(smufl.accSagittal7CommaDown);
note.addModifier('\uE305');
```

Two approaches here:

1) Bundle the most common codes with VexFlow, and distribute the rest in a separate library (maybe `vexflow-ext-smufl`.) For our internal tests, we can pull it in as needed.

2) Depending on how many codes we're talking about, we could simply bundle the whole thing. This should be very compressible. Is there a list somewhere?

tests/accidental_tests.ts
```diff
@@ -286,19 +281,107 @@ function basic(options: TestOptions): void {
   options.assert.ok(true, 'Full Accidental');
 }
 
+function genAccidentals(): string[] {
```
0xfe, line 284: This is a lot nicer!

tests/accidental_tests.ts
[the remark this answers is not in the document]
rvilarl: I do not know if it is complete https://github.com/vexflow/vexflow/blob/main/tools/fonts/config/glyphnames.js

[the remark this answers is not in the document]
0xfe: Thanks Rodrigo. I crunched the numbers on that, and removing the description from each of the codepoints, that data compresses down to ~25k.

The most recent minimized compressed vexflow package is ~213k. So option 2 is about 12% which is pretty sizeable :-(

@ronyeh, what do you think about option 1?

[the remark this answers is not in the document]
rvilarl: There is an option 3. Deliver one with and one without and let the user decide.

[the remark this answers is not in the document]
ronyeh: Welcome back @0xfe ! We have missed your presence & feedback :-)

I'm going to move this conversation over to https://github.com/vexflow/vexflow/issues/23

That thread is now pinned, and will be used to discuss how we can best use SMuFL friendly names in our API.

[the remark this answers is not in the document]
0xfe: Thanks for the kind words, Ron. I've been on a slow recovery, but happy to see all the progress here. Great stuff, everyone!

Will respond back in #23.

rvilarl:
@ronyeh jplease start with this one. This will allow me to do the one for ornaments which is the last one related with glyphs. We will be able to remove all the glyphs files! :) And rethink how to load the web fonts.

ronyeh:
In the current Gonville / Leland examples, the left parentheses seems to touch the flat symbol. The previous version had better spacing (kerning?).

src/accidental.ts
```diff
@@ -50,20 +49,8 @@ export class Accidental extends Modifier {
...
-  protected accidental: {
-    code: string;
-    parenRightPaddingAdjustment: number;
-  };
-  public renderOptions: {
-    parenLeftPadding: number;
-    fontScale: number;
-    parenRightPadding: number;
-  };
+  protected accidental: string;
```
ronyeh, line 52: This is a SMuFL unicode glyph of the form '\uXXXX', correct? Since this approach is new to V5, maybe we should start to label the instance variables where this applies. For example, here we could put a comment:

```ts
// `accidental` is a string from the SMuFL specification.
// For example, '\uE260' is the flat symbol.
// See: `accidentalFlat` at https://w3c.github.io/smufl/latest/tables/standard-accidentals-12-edo.html
```

src/accidental.ts
```diff
@@ -545,94 +522,53 @@ export class Accidental extends Modifier {
...
-        this.renderOptions.parenRightPadding;
-      return this.glyph.getMetrics().width + parenWidth;
+    if (!this.cautionary) {
+      this.text += this.accidental;
+      this.textFont.size = Tables.lookupMetric('Accidental.fontSize');
     } else {
-      return this.glyph.getMetrics().width;
+      this.text += Tables.accidentalCodes('{');
+      this.text += this.accidental;
+      this.text += Tables.accidentalCodes('}');
```
ronyeh, line 533: Why is the shorthand code for left / right parentheses a left / right **brace**? Shouldn't they be `(` and `)` instead?

It looks like this has been around since v3.0.9: https://github.com/0xfe/vexflow/blob/00ec15c67ff333ea49f4d3defbd9e22374c03684/src/accidental.js#L452-L458

@0xfe should the accidental shorthand code for a left/right parenthesis be an actual left/right parenthesis? I suppose we can keep the braces around for backwards compatibility. Just curious about the original reasoning behind this. Or was it because using parentheses makes the code slightly harder to read, since function calls also use parentheses? :-)

src/accidental.ts
```diff
@@ -545,94 +522,53 @@ export class Accidental extends Modifier {
...
+      this.textFont.size = Tables.lookupMetric('Accidental.cautionary.fontSize');
+    }
+    // Accidentals attached to grace notes are rendered smaller.
+    if (isGraceNote(this.note)) {
+      this.textFont.size = 25;
```
ronyeh, line 538: Should this be in lookupMetric as well?

tests/accidental_tests.ts
```diff
@@ -274,9 +272,6 @@ function basic(options: TestOptions): void {
   notes.forEach((note, index) => {
     Note.plotMetrics(f.getContext(), note, 140);
     options.assert.ok(note.getModifiersByType('Accidental').length > 0, 'Note ' + index + ' has accidentals');
-    note.getModifiersByType('Accidental').forEach((accid: Modifier, index: number) => {
-      options.assert.ok(accid.getWidth() > 0, 'Accidental ' + index + ' has set width');
-    });
```
ronyeh, line 279: Just curious why this part of the test is no longer needed?

tests/accidental_tests.ts
```diff
@@ -286,19 +281,107 @@ function basic(options: TestOptions): void {
   options.assert.ok(true, 'Full Accidental');
 }
 
+function genAccidentals(): string[] {
+  const accs = [ '#', '##', 'b', 'bb', 'n', '{', '}', 'db', 'd', '++', '+', '+-'];
+  accs.push('bs', 'bss', 'o', 'k', 'bbs', '++-', 'ashs', 'afhf');
```
ronyeh, line 286: Super minor question. Why do we push a bunch of other items on the second line?
Why not stick it all in the original array? You can always add a // comment if you want to break it up into two lines.

tests/accidental_tests.ts
```diff
@@ -643,39 +714,43 @@ function sagittal(options: TestOptions): void {
   const notes = [
     f
       .StaveNote({ keys: ['d/4', 'f/4', 'b/4', 'b/4'], duration: '4' })
-      .addModifier(accid('accSagittal11MediumDiesisUp'), 1)
-      .addModifier(accid('accSagittal5CommaDown'), 2)
+      .addModifier(accid('\uE30A' /*accSagittal11MediumDiesisUp*/), 1)
+      .addModifier(accid('\uE303' /*accSagittal5CommaDown*/), 2)
```
ronyeh, line 718: Just a note to self: Mohit and Aaron (and possibly others) are in favor of having friendlier ways to insert these unicode strings. So I can look into a good way to allow this.

ronyeh (changes requested):
Mostly small suggestions / questions.

rvilarl:
> In the current Gonville / Leland examples, the left parentheses seems to touch the flat symbol. The previous version had better spacing (kerning?).

Probably kerning/spacing. IT is now rendered as a single string '(S)' where S is the symbol.

src/accidental.ts
rvilarl: I have removed the member variable and used `this.type` directly.

src/accidental.ts
rvilarl: done

tests/accidental_tests.ts
rvilarl: If I put it in one line, prettier will change the arrengement to:
```js
const accs = [ '#', 
    '##', 
    'b',
    ....
```
This increases the number of required lines a lot.

tests/accidental_tests.ts
rvilarl: Back in. I thought that it was not adding value to check that a glyph has a width.

ronyeh:
> 

OK, maybe later on if we want to improve it, we can try out things like '( S )', where we add a space between the parentheses and the symbol?

tests/accidental_tests.ts
ronyeh: OK, this is just in the tests file, so I don't care much. But generally we shouldn't worry about how prettier formats our files. But I've had some luck in adding extra comments // to break up some lines to avoid prettier edge cases. :-)

ronyeh (approved):

PR #39, opened 2023-08-03
Add Academico.

ronyeh:
Added Academico to `WEB_FONT_FILES`, so we can experiment with using it next to Bravura.

In a future PR, we will be more selective with `loadWebFonts()`. Right now it loads ALL the available fonts, and makes the first load very slow (see the developer console network activity when loading `flow.html` for the first time).


rvilarl:
@ronyeh I will merge it but I thought that we were going to delete the binaries from this repository and maintain them in vex-fonts.

ronyeh:
> @ronyeh I will merge it but I thought that we were going to delete the binaries from this repository and maintain them in vex-fonts.

Yes, we will move to a different repository. The new approach needs more testing. In the meantime I just uploaded Academico to the same npm package we were using, so you can test it out as the main text font to pair with Bravura. 

The new approach packages each font as its own npm package. Then when we npm install inside the vexflow repo, it will pull all those fonts into the node_modules directory. Testing can be done all locally once npm install is complete. Currently there is a huge pause when we load flow.html for the first time, since loadWebFonts() downloads all the SMuFL fonts.

...

PR #32, opened 2023-07-26
stavemodifier (v5)

rvilarl:
Now that the extracts have been merged there is only one file changed and no visual differences!

PR #41, opened 2023-08-06
clef (v5) - extracted from stavemodifier (v5)

rvilarl:
@ronyeh following your comment I have divided #32 into smaller parts.

This one is clef only. Visual differences:
- slightly smaller clefs
- `fClef8va/b` & `gClef8va/b` glyphs used directly

Note: adding support for `fClef15ma/b` & `gClef15ma/b` is trivial now.

### DIFF
--- a/src/clef.ts
+++ b/src/clef.ts
@@ -2,31 +2,11 @@
 // MIT License
 // Co-author: Benjamin W. Bohl
 
-import { Glyph } from './glyph';
 import { Stave } from './stave';
 import { StaveModifier, StaveModifierPosition } from './stavemodifier';
 import { Tables } from './tables';
 import { Category } from './typeguard';
-import { defined, log, upperFirst } from './util';
-
-export interface ClefType {
-  code: string;
-  line: number;
-}
-
-export interface ClefAnnotatiomType extends ClefType {
-  xShift: number;
-  point: number;
-}
-
-export interface ClefMetrics {
-  width: number;
-  annotations: {
-    [key: string]: {
-      [type: string]: { line?: number; shiftX?: number } | number;
-    };
-  };
-}
+import { log } from './util';
 
 // eslint-disable-next-line
 function L(...args: any[]) {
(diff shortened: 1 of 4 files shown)


rvilarl:
current
pptr-Clef Clef_Change_Test Bravura svg_current
reference
pptr-Clef Clef_Change_Test Bravura svg_reference

ronyeh:
Just checking in. Should I start reviewing from #41 (this one) onward?

src/clef.ts
```diff
@@ -2,31 +2,11 @@
...
 import { Tables } from './tables';
 import { Category } from './typeguard';
-import { defined, log, upperFirst } from './util';
-
-export interface ClefType {
-  code: string;
-  line: number;
-}
-
-export interface ClefAnnotatiomType extends ClefType {
```
ronyeh, line 17: Ha... good thing we no longer need this interface! It had a typo that we never caught....

ronyeh (changes requested):

ronyeh:
I think the little 8s are a bit too small in the new version... unfortunately.

What do you think about the ligature versions (e.g., gClef8Above):
https://w3c.github.io/smufl/latest/tables/clefs.html?highlight=fClef8va#recommended-ligatures

Or do they look too weird? :-)

ronyeh:
image

The SMuFL spec also has 2/3 sized clefs for doing the clef changes in the middle of a score. Right now we just render the regular clef at a 2/3 scale... but I wonder if using the "correct" glyph will give a slightly better look. But we could tackle that in a different PR...

src/clef.ts
```diff
@@ -46,137 +26,111 @@ export class Clef extends StaveModifier {
...
-    this.#glyphCategory = 'clef' + upperFirst(size);
-
-    const musicFont = Tables.currentMusicFont();
+    this.code = Clef.types[type].code;
+    this.line = Clef.types[type].line;
+    if (size === undefined) {
+      this.size = 'default';
+    } else {
+      this.size = size;
+    }
```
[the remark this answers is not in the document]
rvilarl: done

src/clef.ts
rvilarl: You are right!

PR #43, opened 2023-08-06
timesignature (v5)  - extracted from stavemodifier (v5)

rvilarl:
@ronyeh following your comment I have divided https://github.com/vexflow/vexflow/pull/32 into smaller parts.

This one is time signature only. Visual differences:
- small changes in key signatures (more spacing between digits)
- multi measure rest changes


rvilarl:
current
pptr-Accidental Automatic_Accidentals Bravura svg_current
reference
pptr-Accidental Automatic_Accidentals Bravura svg_reference
current
pptr-canvas_MultiMeasureRest Simple_Test Bravura_current
reference
pptr-canvas_MultiMeasureRest Simple_Test Bravura_reference

ronyeh:
Hmm. I think the older version looks better when there are two digit numbers, like 12 or 10 or 11. I guess this is related to kerning again?

rvilarl:
> Hmm. I think the older version looks better when there are two digit numbers, like 12 or 10 or 11. I guess this is related to kerning again?

In the reference the numbers were rendered independently. The current one is a single statement and it is the browser deciding on the space taking into account the font metrics. I do not think is kerning it is the spacing between characters.

ronyeh:
In the screenshot of the current version of MultiMeasureRest, the last 4 measures seem really weird, especially the 9s.

I don't know this notation well, so I cannot tell whether the new version is incorrect or not. But it looks different enough from the reference that we should investigate.

Did we break the alignment of the glyphs somehow?

rvilarl:
> In the screenshot of the current version of MultiMeasureRest, the last 4 measures seem really weird, especially the 9s.
> 
> I don't know this notation well, so I cannot tell whether the new version is incorrect or not. But it looks different enough from the reference that we should investigate.
> 
> Did we break the alignment of the glyphs somehow?

@mscuthbert @AaronDavidNewman can you help here? Looking at the logic I think that the reference is the faulty one. Do you know?

mscuthbert:
> > In the screenshot of the current version of MultiMeasureRest, the last 4 measures seem really weird, especially the 9s.
> > I don't know this notation well, so I cannot tell whether the new version is incorrect or not. But it looks different enough from the reference that we should investigate.
> > Did we break the alignment of the glyphs somehow?
> 
> @mscuthbert @AaronDavidNewman can you help here? Looking at the logic I think that the reference is the faulty one. Do you know?

If you're talking about the size or kerning of the numbers, I don't have a preference.  If you're talking about the symbols for rests 1-11, they all look fine to me.  The first 9 of the 3rd line on the reference has the glyphs drawn behind the staff lines -- fixed in the current.

(full space rest = breve rest = 2 whole rests;  two-space rests = longa rest = 4 whole rests; modern editions don't use 3 or 4 stave rests (6 or 8 whole rests) so fine that none of these have them)

ronyeh:
Thanks! As long as the current version is the correct one, we're fine. I think the current numbers (at least the last few) are a bit large for my taste...But we can always adjust it in a future PR.

src/multimeasurerest.ts
```diff
@@ -91,34 +74,38 @@ export class MultiMeasureRest extends Element {
...
     super();
+    const fontSize = options.numberGlyphPoint ?? Tables.lookupMetric('MultiMeasureRest.fontSize'); // same as TimeSignature.
+    this.textFont.size = fontSize;
 
     this.numberOfMeasures = numberOfMeasures;
+    this.text = '';
+    const text = `${this.numberOfMeasures}`;
+    for (let i = 0; i < text.length; i++) {
+      this.text += String.fromCodePoint(0xe080 + text.charCodeAt(i) - '0'.charCodeAt(0)) /*timeSigN*/;
+    }
```
ronyeh, line 85: Can this be rewritten as:

```ts
const t = `${this.numberOfMeasures}`;
for (let digit of t) {
  // 0xe080 is timeSig0. We calculate the code point for timeSigN to assemble the digits via SMuFL glyphs.
  this.text += String.fromCodePoint(0xe080 + Number(digit));
}
```

Let me know if it works!

src/multimeasurerest.ts
```diff
@@ -145,36 +132,19 @@ export class MultiMeasureRest extends Element {
...
-    if (this.#hasLineThickness) {
-      lineThicknessHalf = options.lineThickness * 0.5;
-    } else {
-      lineThicknessHalf = spacingBetweenLines * 0.25;
+    let txt = '\ue4ef'; /*restHBarLeft*/
+    const el = new Element();
+    el.setText(txt);
+    el.measureText();
+    for (let i = 1; (i + 2) * el.getWidth() + left <= right; i++) {
+      txt += '\ue4f0'; /*restHBarMiddle*/
```
ronyeh, line 140: This could use some comment right before the for loop. The math is not obvious at first glance.

src/timesignature.ts
```diff
@@ -46,91 +46,98 @@ export class TimeSignature extends StaveModifier {
...
+        code = '\ue090' /*timeSigMinus*/;
+        break;
+      case '(':
+        code = smallSig ? '\ue092' /*timeSigParensLeftSmall*/ : '\ue094' /*timeSigParensLeft*/;
+        break;
+      case ')':
+        code = smallSig ? '\ue093' /*timeSigParensRightSmall*/ : '\ue095' /*timeSigParensRight*/;
+        break;
+      default:
+        code = String.fromCodePoint(0xe080 + key.charCodeAt(0) - '0'.charCodeAt(0)) /*timeSigN*/;
```
ronyeh, line 99: Can it be written like this?

```js
code = String.fromCodePoint(0xe080 + Number(key[0])) /*timeSigN*/;
```

ronyeh (changes requested):

src/multimeasurerest.ts
rvilarl: it works

src/multimeasurerest.ts
rvilarl: done

src/timesignature.ts
rvilarl: done

rvilarl:
Thanks! I will go now to the next step to notes!

PR #44, opened 2023-08-06
musejazz hack to show progress

rvilarl:
This is a hack to show the progress of the migration. Glyph.ts returns always a minus symbol. Only the elements ported are visble. i.e.: StaveTempos

pptr-Stave Tempo_Test MuseJazz svg


rvilarl:
I will also merge this one. It helps me to identify pending elements and it will disappear once the porting is complete.

PR #46, opened 2023-08-07
remove font targets

rvilarl:
Font targets removed, no visual differences


rvilarl:
I will merge this one, font specific builds are no longer required with vexflow 5 and there are no visual changes.

ronyeh:
Thanks. 

Can you describe how we accomplish the same thing with the current build? I want to make sure we didn't remove functionality here. 

For example, a developer wants the minimal vexflow library with only Gonville (i.e. Simon's use case). What does he have to do?

If we call the default loadWebFonts() function, it currently loads a LOT of unnecessary fonts. I assume we should add methods for only loading GonvilleSmufl, etc?

The total network load should be similar to the VexFlow 4.x Gonville only build.

rvilarl:
Hi @ronyeh, I think that we should leave it up to the user to load the fonts, either from internet or their personal web server. I believe that embeding fonts is not a good idea. As @mscuthbert said, the same fonts can be used from different apps or parts of the app. Were you planning to leave them embeded in the library? Do you think that this would have a benefit?

ronyeh:
@rvilarl Yes, we are in agreement. I do not believe we should be "embedding" the OTF into the library.

What I meant is that we need to make it super duper easy for a developer to use VexFlow. 

For example, if a developer wants to use V5 and ONLY load GonvilleSmufl, they should not have to load all the other web fonts that are included in the `loadWebFonts()` method.

What would be the best process for a developer to do this? We need to document this. Ideally, it could look something like this:


```
// ... include the vexflow.js in a script tag ... this library is very small and contains zero fonts.

// This line loads the GonvilleSmufl font from the jsDelivr CDN. No other fonts are loaded.
await Vex.Flow.Font.loadWebFont('Gonville');

// Set the default music engraving font.
Vex.Flow.setMusicFont('Gonville');

// Do other stuff....
....
```

Maybe the setMusicFont() and loadWebFont() should be combined into a single line....?


Alternatively, if the developer wants to pre-load their preferred font from their personal server, they can use the CSS `@font-face` rule to load their own WOFF2 file, before initializing VexFlow.

Anyways, I think we need to document the easiest path to get VexFlow working on a completely clean installation (assuming the developer &amp; end user both have no music fonts on their system).

Issue #38, opened 2023-08-02
Large-scale performance and formatting tests


AaronDavidNewman:
I've been enhancing the reverse-engineering capabilities of Smoosic, so that it can output vex code that runs independently of any application (including Smoosic), and with different versions of vexflow.  I think it fills a void in our testing capabilities , both in terms of scale, and also in the sense that our unit tests aren't very realistic, musically (100s of clefs, etc.)

I don't think these should go in the vexflow repository.  It will be too large - each example is 1000s  of lines and there will be 100-ish different tests.  I think it makes sense to have a different repository that can interact with different versions of vexflow, and also create a useful (if not realistic) performance comparison.

Anyway, here are a couple of examples. 
https://jsfiddle.net/AaronDavidNewman/eoucfL16/

https://jsfiddle.net/AaronDavidNewman/ykx0h76a/7/

sschmidTU:
Great idea!
Unrelated comment: I would y-align lyrics (set each to the lowest y-position in the staffline), otherwise they're unpleasant to read.

rvilarl:
I think also that it is a good idea. What is the source of the text musicxml?

AaronDavidNewman:
> I think also that it is a good idea. What is the source of the text musicxml?

Most of these are either my arrangements, transcriptions, or things I've manually entered in Smoosic to test it out (e.g. GFH Messiah).

I can import music xml (and midi) if you have specific things you think should be represented.  Smoosic has its own serializable format (that's the SMO part of the name).

rvilarl:
@AaronDavidNewman there was a discussion about making tests with larger scores https://github.com/0xfe/vexflow/issues/441 perhaps you could provide some of those.

AaronDavidNewman:
Here is what I have so far.  There is some Joplin, some Bach, some stuff with lyrics, voices, ornaments etc.  Going to add some  Handel and some more instrumental stuff.   At the current time there are 15-20 different vex files, and  6 versions of the vexflow library.

Web interface is very "web 1.0" right now.  Need to add some exception handling and error reporting and I'll add some QUnit-style css.  

When it is better we can fork it into a vexflow project page.

There are some rendering issues:

* 4.0.3 doesn't render anything that uses Leland font.
* Almost nothing renders with the early version of vexflow 5.  Errors in annotation, chord changes and ornaments.
* Some more minor issues with beaming/slur directions that happen different than in Smoosic, not quite sure why yet.

Project web page:
[...]

PR #47, opened 2023-08-07
frethandfinger (v5)

rvilarl:
Minor visual changes

pptr-StringNumber Complex_Measure_With_String___Finger_Numbers Bravura svg_current
pptr-StringNumber Complex_Measure_With_String___Finger_Numbers Bravura svg_reference

### DIFF
--- a/src/frethandfinger.ts
+++ b/src/frethandfinger.ts
@@ -7,7 +7,6 @@ import { Modifier, ModifierPosition } from './modifier';
 import { ModifierContextState } from './modifiercontext';
 import { StemmableNote } from './stemmablenote';
 import { Tables } from './tables';
-import { TextFormatter } from './textformatter';
 import { Category } from './typeguard';
 import { RuntimeError } from './util';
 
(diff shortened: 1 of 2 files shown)


src/frethandfinger.ts
```diff
@@ -117,27 +115,26 @@ export class FretHandFinger extends Modifier {
...
+    this.setFretHandFinger(finger);
     this.position = Modifier.Position.LEFT; // Default position above stem or note head
     this.xOffset = 0; // Horizontal offset from default
     this.yOffset = 0; // Vertical offset from default
   }
 
   setFretHandFinger(finger: string): this {
-    this.finger = finger;
+    this.text = finger;
+    this.measureText();
```
ronyeh, line 132: I still find it a bit confusing that we need to remember to call this.measureText() whenever we change the this.text.

Is it a big performance hit if we automatically update the measurements? I guess maybe there are situations where we want to keep updating the text and batch up all the changes before we call this.measureText() a single time?

src/tables.ts
```diff
@@ -41,7 +41,6 @@ export const CommonMetrics: Record<string, any> = {
   },
 
   FretHandFinger: {
-    fontFamily: 'Arial, sans-serif',
```
ronyeh, line 44: Since fontFamily is not specified here, does that mean that it will default to whatever "text font" we are using? It looks like the new images are using Roboto Slab?

src/frethandfinger.ts
rvilarl: We definetively have to improve on that. #38 will help identifying the performace drivers.

src/tables.ts
rvilarl: They default to the one set in the HACK where you placed the HACK begin / end.

rvilarl:
Thanks!

PR #48, opened 2023-08-08
glyphnote (v5)

rvilarl:
GlyphNote and RepeatNote with minimal visual changes:

current
pptr-GlyphNote GlyphNote_RepeatNote Petaluma svg_current
reference
pptr-GlyphNote GlyphNote_RepeatNote Petaluma svg_reference

### DIFF
--- a/tests/glyphnote_tests.ts
+++ b/tests/glyphnote_tests.ts
@@ -6,7 +6,6 @@
 import { TestOptions, VexFlowTests } from './vexflow_test_helpers';
 
 import { ChordSymbol } from '../src/chordsymbol';
-import { Glyph } from '../src/glyph';
 import { Note } from '../src/note';
 import { Registry } from '../src/registry';
 import { StaveConnector } from '../src/staveconnector';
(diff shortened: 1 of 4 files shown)


rvilarl:
One thing to consider is if RepeatNote is required. Even the tests that we have present repeat notes using GlyphNote directly.

tests/glyphnote_tests.ts
```diff
@@ -38,10 +37,10 @@ function chordChanges(options: TestOptions): void {
...
 
   const notes = [
-    f.GlyphNote(new Glyph('repeatBarSlash', 40), { duration: 'q' }),
-    f.GlyphNote(new Glyph('repeatBarSlash', 40), { duration: 'q' }),
-    f.GlyphNote(new Glyph('repeatBarSlash', 40), { duration: 'q' }),
-    f.GlyphNote(new Glyph('repeatBarSlash', 40), { duration: 'q' }),
+    f.GlyphNote('\uE504' /*repeatBarSlash*/, { duration: 'q' }),
+    f.GlyphNote('\uE504' /*repeatBarSlash*/, { duration: 'q' }),
+    f.GlyphNote('\uE504' /*repeatBarSlash*/, { duration: 'q' }),
+    f.GlyphNote('\uE504' /*repeatBarSlash*/, { duration: 'q' }),
```
ronyeh, line 43: It's interesting that we now ask developers to pass in the unicode code points directly. If not for the comment, this example would be really confusing and unreadable. :-)

We really need to investigate the const enum approach where we could use SMuFL codes directly instead of unicode code points. `repeatBarSlash` is much more readable than `\uE504`.

However, it might inflate the size of our library a bit if we include a regular enum for all the SMuFL code points.

If we can figure out how to get esbuild / webpack to do the right thing, perhaps we can have a separate file that just includes a SMuFL const enum, or even individually exported strings... so we can get the "friendly names" magically replaced by the code points during compilation.

rvilarl:
@ronyeh is the one above the only comment? Can this be merged?

tests/glyphnote_tests.ts
ronyeh: Yes, we would not include the entire SMuFL list if it expands the library.

I would either:
1) use a `const enum` but verify that esbuild produces a library where all the enums are replaced by their code points. The const enum should have zero impact on the final size of the library.

or 

2) use a regular enum but with just the subset of glyphs that we use internally in VexFlow.

or

3) export individual strings from a SMuFL module, but with only a subset of the glyphs that we use internally in VexFlow.

ronyeh:
Thanks!

PR #50, opened 2023-08-08
stavesection (v5)

rvilarl:
Minor changes, rect is better aligned.

current
pptr-Stave Multiple_Stave_Barline_Test Bravura svg_current
reference
pptr-Stave Multiple_Stave_Barline_Test Bravura svg_reference

### DIFF
--- a/src/stavesection.ts
+++ b/src/stavesection.ts
@@ -3,7 +3,6 @@
 
 import { Stave } from './stave';
 import { StaveModifier } from './stavemodifier';
-import { TextFormatter } from './textformatter';
 import { Category } from './typeguard';
 
 export class StaveSection extends StaveModifier {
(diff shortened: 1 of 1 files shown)


src/stavesection.ts
```diff
@@ -3,72 +3,52 @@
...
     return Category.StaveSection;
   }
 
-  protected section: string;
-  protected shiftX: number;
-  protected shiftY: number;
   protected drawRect: boolean;
 
-  constructor(section: string, x: number, shiftY: number, drawRect = true) {
+  constructor(section: string, x: number, yShift: number, drawRect = true) {
```
ronyeh, line 15: Interesting. Some classes use shiftY and other classes use yShift.

Should we standardize? Which one is better? Do we also use deltaX, dX or something similar?

rvilarl: I am standarizing to `yShift` as already used in the base class `Element`

ronyeh:
This can be a separate PR, but I think the rect needs more internal padding all around. Right now it crowds the letter B in all 4 directions. :-)

PR #51, opened 2023-08-09
annotation (v5)

rvilarl:
Annotations follow now the specified font. Several layout improvements as well.

Some visual changes with Petluma.

current
pptr-Annotation Lyrics Petaluma svg_current
reference
pptr-Annotation Lyrics Petaluma svg_reference

current
pptr-Annotation Bottom_Annotations_with_Beams Petaluma svg_current
reference
pptr-Annotation Bottom_Annotations_with_Beams Petaluma svg_reference

current
pptr-Beam Complex_Beams_with_Annotations Petaluma svg_current
reference
pptr-Beam Complex_Beams_with_Annotations Petaluma svg_reference

### DIFF
--- a/tests/annotation_tests.ts
+++ b/tests/annotation_tests.ts
@@ -15,7 +15,6 @@ import { Beam } from '../src/beam';
 import { Bend } from '../src/bend';
 import { ElementStyle } from '../src/element';
 import { Flow } from '../src/flow';
-import { Font, FontStyle, FontWeight } from '../src/font';
 import { Formatter } from '../src/formatter';
 import { ModifierPosition } from '../src/modifier';
 import { Registry } from '../src/registry';
@@ -23,6 +22,7 @@ import { ContextBuilder } from '../src/renderer';
 import { Stave } from '../src/stave';
 import { StaveNote, StaveNoteStruct } from '../src/stavenote';
 import { Stem } from '../src/stem';
+import { Tables } from '../src/tables';
 import { TabNote, TabNoteStruct } from '../src/tabnote';
 import { TabStave } from '../src/tabstave';
 import { Tickable } from '../src/tickable';
@@ -84,7 +84,7 @@ function lyrics(options: TestOptions): void {
       const verse = Math.floor(ix / 3);
       const noteGroupID = 'n' + (ix % 3);
       const noteGroup = registry.getElementById(noteGroupID) as Tickable;
-      const lyricsAnnotation = f.Annotation({ text }).setFont('Roboto Slab', fontSize);
+      const lyricsAnnotation = f.Annotation({ text }).setFontSize(fontSize);
       lyricsAnnotation.setPosition(ModifierPosition.BELOW);
       noteGroup.addModifier(lyricsAnnotation, verse);
     });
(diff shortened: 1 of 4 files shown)


tests/beam_tests.ts
```diff
@@ -771,7 +772,7 @@ function complexWithAnnotation(options: TestOptions): void {
   ];
 
   const font = {
-    family: Font.SERIF,
+    family: Tables.lookupMetric('Annotation.fontFamily'),
```
ronyeh, line 775: Is Tables a public class? I always thought it was for internal calculations.

I think the files under `tests/*_tests.ts` should be representative of how a developer would use VexFlow. So if a developer can't easily access Tables, we shouldn't import it and should not use it here.

tests/annotation_tests.ts
```diff
@@ -195,11 +201,11 @@ function harmonic(options: TestOptions, contextBuilder: ContextBuilder): void {
...
   const ctx = contextBuilder(options.elementId, 500, 240);
 
-  ctx.setFont(Font.SANS_SERIF, FONT_SIZE);
-  const stave = new TabStave(10, 10, 450).addTabGlyph().setContext(ctx).draw();
+  ctx.setFont(Tables.lookupMetric('fontFamily'), FONT_SIZE);
+  const stave = new TabStave(10, 10, 450).addClef('tab').setContext(ctx).draw();
 
   const annotation = (text: string) =>
-    new Annotation(text).setFont(Font.SERIF, FONT_SIZE, FontWeight.NORMAL, FontStyle.ITALIC);
+    new Annotation(text).setFont(Tables.lookupMetric('Annotation.fontFamily'), FONT_SIZE, 'normal', 'italic');
```
ronyeh, line 208: Same concern here. I think a developer using the Annotation class shouldn't have to call `Tables.lookupMetric('Annotation.fontFamily')`.

Shouldn't they just call `new Annotation(text)`? Should the font default to the one they chose for the text font (presumably earlier)?

Basically when a developer adds a new VexFlow score on a webpage, I imagine they should do two things:

1) set the music engraving font, and load the font file
2) set the text font, and load the font file if needed
3) ... everything else.

I've been looking at the SVG elements output by VexFlow 5 and I find it weird that the music symbol elements have fontFamily: "Bravura, Roboto Slab". I probably will file an issue for this... so we can investigate. 

I think long term, elements that are only musical glyphs should only have a music font stack. Elements that only have text glyphs should only have a text font stack (e.g., "Academico, Times New Roman, serif").

Elements that are for some reason mixed (both music and text symbols) are allowed to have a mixed font stack (e.g, "Bravura, Academico").

tests/annotation_tests.ts
```diff
@@ -15,14 +15,14 @@ import { Beam } from '../src/beam';
...
 import { Flow } from '../src/flow';
-import { Font, FontStyle, FontWeight } from '../src/font';
 import { Formatter } from '../src/formatter';
 import { ModifierPosition } from '../src/modifier';
 import { Registry } from '../src/registry';
 import { ContextBuilder } from '../src/renderer';
 import { Stave } from '../src/stave';
 import { StaveNote, StaveNoteStruct } from '../src/stavenote';
 import { Stem } from '../src/stem';
+import { Tables } from '../src/tables';
```
ronyeh, line 25: I know a couple of the test files import Tables, but did we ever decide of Tables is supposed to be internal use, or for public use?

If it is for public use, then we should be much more careful with the design of the API. Right now, Tables is sort of a kitchen sink of useful object dictionaries that are used by our internal classes.

I think Tables was intended to be internal, so that's why we had a `flow.ts` class that made some of the internal Tables data public.

So we should be careful when we directly access Tables in our tests, because the test code is how some developers learn how to use the public APIs.

tests/beam_tests.ts
rvilarl: @ronyeh lookupMetric should be public. I suggest to leave the solution out of this PR. Perhpas tables is not the right place or we need and access function from Vex.Flow.

tests/annotation_tests.ts
rvilarl: same us above

tests/annotation_tests.ts
rvilarl: same as above, please leave it to another PR

tests/annotation_tests.ts
ronyeh: OK, we will revisit this font stack question once more parts are migrated.

Issue #49, opened 2023-08-08
Error in stavehairpin.ts


AaronDavidNewman:
I get the following exception.  Prior versions of vexflow had `notes.first_note `

``` typescript
  const smo390376 = new VF.StaveHairpin({ first_note: smo362522, last_note: smo362522 });
...
    setNotes(notes) {
        if (!notes.firstNote && !notes.lastNote) {
            throw new RuntimeError('BadArguments', 'Hairpin needs to have either firstNote or lastNote set.');
        }
```

rvilarl:
Try

`const smo390376 = new VF.StaveHairpin({ firstNote: smo362522, lastNote: smo362522 }); ...`

AaronDavidNewman:
Would this work for 4.x libraries also?  I'm trying to create code that will run on different libraries for comparison purposes.

rvilarl:
No, there was the decision to change to camel case. In 4.x camel case amd snake case are mixed

AaronDavidNewman:
I didn't realize this decision meant we were breaking every interface.  The old code was pretty consistent, in the sense that shift_x in render_options was always notated that way.  We didn't even add a type that we could use at compile time to determine if there was an issue, it is still just a `Record<string, note>`, and you need to look at the code to know what the right string is.

What would be wrong with:

``` typescript
  setNotes(notes: Record<string, Note>): this {
if (notes['first_note]') notes.firstNote = notes.first_note; // backwards compatibility
```

30 seconds of typing, saves hours possible for a user.  No run-time penalty, you still get the refactoring you want.  win/win.

ronyeh:
Thanks for this discussion.

We will definitely need to figure out [ whether to / how to best  ] support backwards compatibility with v4.

When we went to camelCase, we broke a lot of APIs and we figured it was OK since this is a major major new revision of VexFlow.

However, it you think it will cause people to never want to upgrade (because it's too painful) then we should figure out a good way to support the old APIs with underscores in the strings.

In any case, keep on reporting new issues as you discover them! This will help make the V5 as good a product as possible.

Will you be able to share a landing page where you compare V4 with V5 (assuming we can figure these issues out)?

@rvilarl I think this raises a good question of how much we are breaking during this migration. Will your OSMD / MusicXML app support both V4 and V5? Or will there be an easy way to see both outputs in different tabs so we can flip between them to see the visual difference in the scores?

[...]

PR #54, opened 2023-08-10
Add ornament codes for backwards compatibility

AaronDavidNewman:
Since these are hard-coded strings and have been around for 10 years.

### DIFF
--- a/src/tables.ts
+++ b/src/tables.ts
@@ -662,8 +662,10 @@ const articulations: Record<string, ArticulationStruct> = {
 const ornaments: Record<string, { code: string }> = {
   mordent: { code: 'ornamentShortTrill' },
   mordentInverted: { code: 'ornamentMordent' },
+  mordent_inverted: { code: 'ornamentMordent' }, // backwards-compatible to 4.x
   turn: { code: 'ornamentTurn' },
   turnInverted: { code: 'ornamentTurnSlash' },
+  turn_inverted: { code: 'ornamentTurnSlash' }, // backwards-compatible to 4.x
   tr: { code: 'ornamentTrill' },
   upprall: { code: 'ornamentPrecompSlideTrillDAnglebert' },
   downprall: { code: 'ornamentPrecompDoubleCadenceUpperPrefix' },


rvilarl:
@AaronDavidNewman The plan was to be disrupted with snake case, but I am flexible.
@ronyeh @mscuthbert what do you think?

This is also related to #49 

We could also go for a general function applied to parameters like:
```
function camelToUnderscore(key) {
   var result = key.replace( /([A-Z])/g, " $1" );
   return result.split(' ').join('_').toLowerCase();
}
```

AaronDavidNewman:
I would argue that the ornament codes are different than the API.  There is no way to create ornaments except by passing in the specific code, and no way to programmatically detect what the right code is, since the data in Tables isn't exported.  

When I added the jazz ornaments to the ornaments class, I went to great lengths to make sure the existing ornaments still worked the same way, out of respect for the existing vexflow user base (and because the other developers asked me to), even though it was sort of a pain to have some ornament codes in tables and some in the ornaments class.  We should try to make the code backwards compatible when we can, in this case there is no penalty for doing so (we have the new format and the old format).

ronyeh:
First off, here's my initial opinion :-)

I'd prefer to not have duplicate entries to support backwards compatibility (since V5 is breaking almost every API anyways). Is it OK if we just document this change, and ask the developer to pass in a camelCased key for the ornament code?

But if this is a big burden on developers, then I'm also flexible....

* * * 

Are `turn_inverted` and `mordent_inverted` the only two codes that use underscores? Will we have to add more backwards compatibility for other ornament codes as well?

I see other camelCased and multiword keys in ornamentCodes. I want to make sure we don't need to also "fix" the other keys like `jazzTurn` or `plungerClosed` or `downmordent` someday.

Or what if.... we just make all the keys **lowercase with no underscores**?

```
const ornaments: Record<string, { code: string }> = {
  mordent: { code: 'ornamentShortTrill' },
  mordentinverted: { code: 'ornamentMordent' },
  turn: { code: 'ornamentTurn' },
  turninverted: { code: 'ornamentTurnSlash' },
  tr: { code: 'ornamentTrill' },
  upprall: { code: 'ornamentPrecompSlideTrillDAnglebert' },
  downprall: { code: 'ornamentPrecompDoubleCadenceUpperPrefix' },
  prallup: { code: 'ornamentPrecompTrillSuffixDandrieu' },
  pralldown: { code: 'ornamentPrecompTrillLowerSuffix' },
  upmordent: { code: 'ornamentPrecompSlideTrillBach' },
  downmordent: { code: 'ornamentPrecompDoubleCadenceUpperPrefixTurn' },
  lineprall: { code: 'ornamentPrecompAppoggTrill' },
  prallprall: { code: 'ornamentTremblement' },
  scoop: { code: 'brassScoop' },
  doit: { code: 'brassDoitMedium' },
  fall: { code: 'brassFallLipShort' },
  doitlong: { code: 'brassLiftMedium' },
  falllong: { code: 'brassFallRoughMedium' },
  bend: { code: 'brassBend' },
  plungerclosed: { code: 'brassMuteClosed' },
  plungeropen: { code: 'brassMuteOpen' },
  flip: { code: 'brassFlip' },
  jazzturn: { code: 'brassJazzTurn' },
  smear: { code: 'brassSmear' },
};
```

Then the Ornament constructor can do this:

```ts
  constructor(type: string) {
    super();

    // Remove undescrores and conver to lower case, for compatibility with V4.
    type = type.replace(/_/g, '').toLowerCase();

    this.type = type;
```


Would this work for everyone? It would be the most "flexible" solution since it would support both camelCased keys and keys with underscores.

AaronDavidNewman:
I think if you're going to change the API, you should expose the changes in the API.  The problem with the ornament codes is that they are hard-coded strings in a file that isn't publicly accessible outside of the library.   They are only types inside of Tables.ts.

I can get around the other compatibility issues by passing in both sets of parameters.  I.e. I will have to make my changes backwards compatible.  But in this case it isn't even possible for me to do so.

ronyeh:
Thanks for the context. Since you joined this project way before Rodrigo or I did, we'll defer to you on this matter of backwards compatibility.

One of our goals for V5 is to "clean up" and modernize the library. Since it's a major new revision, we are OK with breaking backwards compatibility.

If we were designing this class from scratch, perhaps we would use an enum that contains all the different OrnamentTypes? That would improve the type safety a little, instead of us just accepting arbitrary string codes (which are then internally remapped to a different string code for SMuFL purposes).

Maybe in Rodrigo's @rvilarl conversion of the Ornament class, we'll skip the  SMuFL glyph names and go directly to the unicode code points?

For example, we can probably skip this translation step:
```
downmordent: { code: 'ornamentPrecompDoubleCadenceUpperPrefixTurn' },
```

And instead do something like this.

```
downmordent: { code: '\uE5C4' },
```

Also... is there a good reason to keep `ornamentCodes` in the Tables class? Could we pull it out into an `OrnamentType` file that is just a string enum? I'm not a huge fan of how our tables.ts has become the kitchen sink of useful mappings for lots of different classes :-)....

ronyeh:
I'm merging this one for now, but will leave open the possibility that we may discover a cleaner approach for specifying Ornament types.

PR #56, opened 2023-08-12
tremolo (v5)

rvilarl:
Big tremolos removed, they are big enough when following the SMuFL ratio. x & y offsets calculated at run time based on the note position. Y position calculated from end of stem.

current
pptr-Tremolo Tremolo___Basic Bravura svg_current
reference
pptr-Tremolo Tremolo___Basic Bravura svg_reference

### DIFF
--- a/src/fonts/common_metrics.ts
+++ b/src/fonts/common_metrics.ts
@@ -156,23 +156,6 @@ export const CommonMetrics = {
     },
   },
 
-  tremolo: {
-    default: {
-      spacing: 7,
-      offsetYStemUp: -8,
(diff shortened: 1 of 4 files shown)


ronyeh:
I see we still use a for loop to add up to `this.num` tremolo slashes.

Will we consider using the `tremolo2` `tremolo3` `tremolo4` &amp; `tremolo5` glyphs? Presumably they are spaced appropriately (the way the font designer wants them to be spaced).

https://w3c.github.io/smufl/latest/tables/tremolos.html?highlight=tremolo#tremolos-ue220ue23f

PR #57, opened 2023-08-14
textnote (v5)

rvilarl:
TextNote and TextDynamics ported.

### DIFF
--- a/src/textdynamics.ts
+++ b/src/textdynamics.ts
@@ -1,7 +1,6 @@
 // Copyright (c) 2023-present VexFlow contributors: https://github.com/vexflow/vexflow/graphs/contributors
 // MIT License
 
-import { Glyph } from './glyph';
 import { Note } from './note';
 import { Tables } from './tables';
 import { TextNoteStruct } from './textnote';
@@ -31,35 +30,16 @@ export class TextDynamics extends Note {
   protected sequence: string;
 
   protected line: number;
-  protected glyphs: Glyph[];
 
   /** The glyph data for each dynamics letter. */
-  static get GLYPHS(): Record<string, { code: string; width: number }> {
+  static get GLYPHS(): Record<string, string> {
     return {
-      f: {
-        code: 'dynamicForte',
-        width: 12,
-      },
-      p: {
-        code: 'dynamicPiano',
-        width: 14,
-      },
-      m: {
-        code: 'dynamicMezzo',
-        width: 17,
-      },
-      s: {
-        code: 'dynamicSforzando',
-        width: 10,
-      },
-      z: {
-        code: 'dynamicZ',
-        width: 12,
-      },
-      r: {
-        code: 'dynamicRinforzando',
-        width: 12,
-      },
+      f: '\uE522' /*dynamicForte*/,
+      p: '\uE520' /*dynamicPiano*/,
+      m: '\uE521' /*dynamicMezzo*/,
+      s: '\uE524' /*dynamicSforzando*/,
+      z: '\uE525' /*dynamicZ*/,
+      r: '\uE523' /*dynamicRinforzando*/,
     };
   }
 
@@ -77,10 +57,10 @@ export class TextDynamics extends Note {
 
     this.sequence = (noteStruct.text || '').toLowerCase();
     this.line = noteStruct.line || 0;
-    this.glyphs = [];
+    this.text = '';
 
-    this.renderOptions = { ...this.renderOptions, glyphFontSize: Tables.NOTATION_FONT_SCALE };
-
+    this.renderOptions = { glyphFontSize: Tables.lookupMetric('fontSize'), ...this.renderOptions };
+    this.textFont.size = defined(this.renderOptions.glyphFontSize) * this.renderOptions.glyphFontScale;
     L('New Dynamics Text: ', this.sequence);
   }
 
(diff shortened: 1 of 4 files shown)


rvilarl:
current
pptr-TextNote Text_Dynamics Bravura svg_current
reference
pptr-TextNote Text_Dynamics Bravura svg_reference
current
pptr-TextNote TextNote_Formatting_With_Glyphs_0 Bravura svg_current
reference
pptr-TextNote TextNote_Formatting_With_Glyphs_0 Bravura svg_reference

ronyeh:
I notice how some of the symbols are no longer vertically centered. The last reference image has big symbols that are vertically centered. Is this an intentional change?

src/textdynamics.ts
```diff
@@ -31,35 +30,16 @@ export class TextDynamics extends Note {
...
-      },
-      z: {
-        code: 'dynamicZ',
-        width: 12,
-      },
-      r: {
-        code: 'dynamicRinforzando',
-        width: 12,
-      },
+      f: 'E522' /*dynamicForte*/,
```
rvilarl, line 37: Change to unicode.

src/textnote.ts
```diff
@@ -207,35 +152,21 @@ export class TextNote extends Note {
...
-      if (this.subscript) {
-        ctx.setFont(family, smallerFontSize, weight, style);
-        ctx.fillText(this.subscript, x + this.width + 2, y + height / 2.2 - 1);
-      }
-
-      this.restoreStyle(ctx);
+    const y = stave.getYForLine(this.line + -3);
+    this.applyStyle(ctx);
+    ctx.setFont(this.textFont);
+    ctx.fillText(this.text, x, y);
```
rvilarl, line 158: Use RenderText

rvilarl (commented):
Some comments

rvilarl:
> I notice how some of the symbols are no longer vertically centered. The last reference image has big symbols that are vertically centered. Is this an intentional change?

I have reviewed it and I do not see the reason. It is not intencional.

src/textdynamics.ts
rvilarl: done

src/textnote.ts
rvilarl: done

tests/textnote_tests.ts
```diff
@@ -140,7 +140,7 @@ function superscriptAndSubscript(options: TestOptions): void {
   const voice2 = score.voice(notes2);
   voice2.getTickables().forEach((note) => {
     const textNote = note as TextNote;
-    textNote.setFont({ family: Font.SERIF, size: 15 });
+    textNote.setFont({ family: Tables.lookupMetric('fontFamily'), size: 15 });
```
ronyeh, line 143: I would avoid using Tables.lookupMetric(...) in our *_tests.ts files.

I treat the tests as one way to teach developers how to use the public API. So if we do this here, we are declaring that Tables.lookupMetric('fontFamily') is a public API that should be relied on.

I'd rather just put a string here, like 'serif'.

If we do not specify the fontFamily, won't it default to the right thing?

Can this be written as:

```
 textNote.setFont({ size: 15 });
```

tests/textnote_tests.ts
```diff
@@ -7,9 +7,9 @@ import { TestOptions, VexFlowTests } from './vexflow_test_helpers';
 
 import { Crescendo } from '../src/crescendo';
 import { Flow } from '../src/flow';
-import { Font } from '../src/font';
 import { Note } from '../src/note';
 import { Stave } from '../src/stave';
+import { Tables } from '../src/tables';
```
ronyeh, line 12: Are we suggesting developers to use Tables directly? I think in the past, we have considered it more of an internal class.

src/textnote.ts
```diff
@@ -112,26 +71,29 @@ export class TextNote extends Note {
...
 
     this.text = noteStruct.text || '';
-    this.superscript = noteStruct.superscript;
-    this.subscript = noteStruct.subscript;
-    this.setFont(noteStruct.font);
+    if (noteStruct.glyph) {
+      this.text += TextNote.GLYPHS[noteStruct.glyph] || noteStruct.glyph;
+    }
+    if (noteStruct.font) this.setFont(noteStruct.font);
+    else if (noteStruct.glyph == undefined) this.setFont(Tables.lookupMetricFontInfo('TextNote.text.fontSize'));
```
ronyeh, line 78: We should try to avoid double equals == in our code.

Here we can use `noteStruct.glyph === undefined` or just write `!noteStruct.glyph` if you like.

ronyeh: Also, for this if/else if/ else, I'd prefer to use braces { } even though each case is a single line. It makes it easier to read/skim in my opinion.

ronyeh (changes requested):
OK mostly looks good. Just some small suggestions.

tests/textnote_tests.ts
rvilarl: I changed it to setFontSize

tests/textnote_tests.ts
rvilarl: Done

src/textnote.ts
rvilarl: done

rvilarl:
@ronyeh ready

src/textnote.ts
```diff
@@ -112,26 +71,33 @@ export class TextNote extends Note {
...
+    }
+    if (noteStruct.font) {
+      this.setFont(noteStruct.font);
+    } else if (noteStruct.glyph === undefined) {
+      this.setFont(Tables.lookupMetricFontInfo('TextNote.text.fontSize'));
+    } else {
+      this.measureText();
+    }
+    // Scale the font size by 1/1.3.
+    const smallerFontSize = Font.convertSizeToPointValue(this.textFont.size) * 0.769231;
```
ronyeh, line 85: Maybe someday we can make this 0.769231 customizable. For example, it could be in the Metrics class (aka CommonMetrics).

ronyeh (approved):

PR #61, opened 2023-08-18
Loading fonts from node_modules/

ronyeh:
The big idea is that all the fonts are npm installed into node_modules.

Then our `flow.html` and glyph inspector tools will load fonts directly from `node_modules` via @font-face rules in CSS files.

This will enable offline development (e.g., when you are on an airplane), as discussed a while back.

At some point, we need to figure out how to teach people to load the correct web font that they desire, so their deployment of VexFlow doesn't show empty squares / "tofu blocks".


ronyeh:
Also up for consideration....

Maybe we should make a subset version of Bravura called `BravuraVF` that only includes the common SMuFL code points that developers tend to use. We can start with the ones that Leland supports, and if anyone complains about a missing glyph, we can add it to `BravuraVF`.

For comparison:

Bravura.woff2 is 247KB while Leland.woff2 is 36KB.

It'll make websites load faster especially since scores don't have a good fallback font (that is preinstalled on OSes), so we have to block rendering until the font is completely loaded.

rvilarl:
I would not create subsets. I would leave fonts unchanged. 

The users can always create their own fonts, FontForge is pretty easy to use.

ronyeh:
Opening `flow.html` now works with fonts loaded from node_modules, so it works offline.

**Last thing to investigate**: 
`grunt generate:current` doesn't seem to load the fonts properly for the *.svg.png output (I get empty squares). 

I also get empty squares in the `*.svg` files, but this makes sense, since the fonts are NOT installed locally on my machine. (I am testing a real world situation where the user has zero music fonts on their computer.) IMO this is the biggest weakness of V5 at the moment. It does not make portable SVG files since it uses the glyph code points directly, and not the paths.

@rvilarl Have you been experiencing the same issues with image generation (empty squares with *.svg.png)?

I think it might be because the generate_images_pptr.js script makes an SVG, and then asks puppeteer to convert that SVG to a PNG. However, SVGs cannot load fonts from `node_modules`, so if the user does not have the correct fonts on their OS, they will see empty squares..... :-(

rvilarl:
@ronyeh yes I got this problem at one stage when I did not have the fonts installed.

ronyeh:
This is ready for review. Please test it out (and try turning off your WiFI)!

If you look at `fonts.css`, we make use of the `local()` loader, so if the music fonts are installed on your machine, it won't even bother looking in node_modules.

The SVG issue is pretty big in my opinion. It would be nice to be able to make portable SVGs (maybe if the developer wants to "opt in" to portable SVGs, they would have to npm install opentype.js, so we can read the font paths at runtime).

Anyways, regardless what we end up doing, we need to always keep in mind that most developers (and almost 100% of end users) do NOT have SMuFL fonts installed on their operating systems.

ronyeh:
OK, please give this a go.

Remember you'll need to do an `npm install` after pulling down this branch, because all the webfonts are now individual npm modules.

ronyeh:
A bunch of files were moved into the `vexflow/demos/fonts/old/` directory. At some point, I may delete these if the examples are no long useful.

tests/flow-headless-browser.html
```diff
@@ -1,16 +1,41 @@
...
+      <span style="font-family: GonvilleSmufl"></span>
+      <span style="font-family: Gootville"></span>
+      <span style="font-family: Leland"></span>
+      <span style="font-family: MuseJazz"></span>
+      <span style="font-family: Petaluma"></span>
+      <span style="font-family: Petaluma Script"></span>
+      <span style="font-family: Roboto Slab"
+        ><span style="font-weight: 500"></span><span style="font-weight: 700"></span
+      ></span>
+    </div>
```
ronyeh, line 29: I copied your glyph inspector approach to preloading font files.

Note that for text fonts with multiple styles and weights, you have to add `<span>`, `<b>`, `<i>` children with the appropriate styles, to trigger the correct fonts to download.

tests/flow.html
```diff
@@ -93,36 +160,52 @@ <h3>
...
-        loadVexFlow = () => loadScript(vexPlusTestsURL);
+        loadVexFlow = async () => {
+          await loadScript(vexPlusTestsURL);
+          return Vex.Flow;
+        };
       }
 
-      await loadVexFlow();
+      // Only proceed after all fonts have loaded.
+      const fontFaceSet = await document.fonts.ready;
```
ronyeh, line 187: This is one approach to waiting until all fonts are loaded before starting VexFlow.

The other way is to call `loadWebFonts(...)`, which will load a woff2 from a CDN.

tests/fonts.css
```diff
@@ -0,0 +1,210 @@
...
+/* Bravura */
+@font-face {
+  font-family: 'Bravura';
+  font-style: normal;
+  font-display: block;
+  font-weight: 400;
+  src:
+    local('Bravura'),
+    url(../node_modules/@vexflow-fonts/bravura/bravura.woff2) format('woff2');
+}
```
ronyeh, line 30: `local('Bravura')` asks the browser to use the font installed on the user's operating system, instead of loading from the url().

If your system has these fonts installed, you'll see that the developer tools / network tab never loads the font file.

rvilarl (approved):

rvilarl:
LGTM Thanks Ron!

PR #65, opened 2023-08-20
Use loadWebFonts(fontNames) to load only a few fonts. 

ronyeh:
Omit the `fontNames` parameter to load all the web fonts compatible with VexFlow.

### DIFF
--- a/src/font.ts
+++ b/src/font.ts
@@ -373,22 +373,29 @@ export class Font {
   }
 
   /**
-   * Load the web fonts that are used by ChordSymbol. For example, `flow.html` calls:
+   * Load the web fonts that are used by your app.
+   * If fontNames is undefined, all fonts in Font.WEB_FONT_FILES will be loaded.
+   *
+   * For example, `flow.html` calls:
    *   `await Vex.Flow.Font.loadWebFonts();`
    * Alternatively, you may load web fonts with a stylesheet link (e.g., from Google Fonts),
    * and a @font-face { font-family: ... } rule in your CSS.
-   * If you do not load either of these fonts, ChordSymbol will fall back to Times or Arial,
-   * depending on the current music engraving font.
    *
    * You can customize `Font.WEB_FONT_HOST` and `Font.WEB_FONT_FILES` to load different fonts
    * for your app.
    */
-  static async loadWebFonts(): Promise<void> {
+  static async loadWebFonts(fontNames?: string[]): Promise<void> {
+    const allFiles = Font.WEB_FONT_FILES;
+    if (!fontNames) {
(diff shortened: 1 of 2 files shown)


tests/flow.html
```diff
@@ -74,28 +71,15 @@ <h3>
...
         const version = ver.split('@')[1];
-        if (parseFloat(version) < 4) {
-          isVersionFourOrNewer = false;
-          vexURL = `${cdnURLPath}vexflow@${version}/releases/vexflow-debug.js`;
-          testsURL = `${cdnURLPath}vexflow@${version}/releases/vexflow-tests.js`;
-        } else {
-          // VexFlow 4 moved the build output from releases/ to build/.
-          vexURL = `${cdnURLPath}vexflow@${version}/build/vexflow-debug.js`;
-          vexPlusTestsURL = `${cdnURLPath}vexflow@${version}/build/vexflow-debug-with-tests.js`;
-        }
```
ronyeh, line 86: Remove some legacy code from flow.html. From now on we only support versions >= 4.0.

src/font.ts
```diff
@@ -373,22 +373,29 @@ export class Font {
...
    * Alternatively, you may load web fonts with a stylesheet link (e.g., from Google Fonts),
    * and a @font-face { font-family: ... } rule in your CSS.
-   * If you do not load either of these fonts, ChordSymbol will fall back to Times or Arial,
-   * depending on the current music engraving font.
    *
    * You can customize `Font.WEB_FONT_HOST` and `Font.WEB_FONT_FILES` to load different fonts
    * for your app.
    */
-  static async loadWebFonts(): Promise<void> {
+  static async loadWebFonts(fontNames?: string[]): Promise<void> {
```
ronyeh, line 387: Add an optional parameter to allow loading just a few of the web fonts.

src/font.ts
```diff
@@ -373,22 +373,29 @@ export class Font {
...
 
   /**
-   * Load the web fonts that are used by ChordSymbol. For example, `flow.html` calls:
+   * Load the web fonts that are used by your app.
+   * If fontNames is undefined, all fonts in Font.WEB_FONT_FILES will be loaded.
+   *
+   * For example, `flow.html` calls:
    *   `await Vex.Flow.Font.loadWebFonts();`
    * Alternatively, you may load web fonts with a stylesheet link (e.g., from Google Fonts),
    * and a @font-face { font-family: ... } rule in your CSS.
```
rvilarl, line 382: hi @ronyeh,
the approach I like most for the future is to use font-face statements. I did that in the inspection tool `glyphs.html`. But I do not know how to force an await on css. do you?

rvilarl (approved):

PR #66, opened 2023-08-21
pedalmarking (v5)

rvilarl:
PedalMarking ported with visual differences.

### DIFF
--- a/src/pedalmarking.ts
+++ b/src/pedalmarking.ts
@@ -2,7 +2,6 @@
 // MIT License
 
 import { Element } from './element';
-import { Glyph } from './glyph';
 import { RenderContext } from './rendercontext';
 import { StaveNote } from './stavenote';
 import { Tables } from './tables';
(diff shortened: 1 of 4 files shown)


src/pedalmarking.ts
```diff
@@ -19,11 +18,11 @@ function L(...args: any[]) {
...
  * coordinate shifts.
  */
-function drawPedalGlyph(name: string, context: RenderContext, x: number, y: number, point: number): void {
-  const glyphData = PedalMarking.GLYPHS[name];
-  const glyph = new Glyph(glyphData.code, point, { category: 'pedalMarking' });
-  // Center the middle of the glyph with the middle of the note head (Tables.STAVE_LINE_DISTANCE / 2)
-  glyph.render(context, x - (glyph.getMetrics().width - Tables.STAVE_LINE_DISTANCE) / 2, y);
+function drawPedalGlyph(name: string, ctx: RenderContext, x: number, y: number): void {
+  const glyph = new Element(PedalMarking.CATEGORY);
+  glyph.setText(PedalMarking.GLYPHS[name] || name);
```
ronyeh, line 23: Effectively the same, but I think we should use `??` operator when we are checking for undefined / null. We should keep `||` for actual boolean expressions (e.g., in an if / else if condition).

ronyeh:
Yes, we might as well change the test case to use normal weight. Bold looks too heavy. Thanks for catching the bug!

ronyeh:
OK, I can merge this after the two small changes.

src/pedalmarking.ts
rvilarl: done

rvilarl:
current
pptr-PedalMarking Custom_Text_2 Bravura svg_current
reference
pptr-PedalMarking Custom_Text_2 Bravura svg_reference

@ronyeh ready to review,/merge. Now that I see  that, the reference was probably bold.

PR #67, opened 2023-08-22
element x y integration, extracted from Note (v5).

rvilarl:
No visual diferencies, once merged I will rebase #34 and the number of modified files will decrease.

### DIFF
--- a/src/clefnote.ts
+++ b/src/clefnote.ts
@@ -45,6 +45,6 @@ export class ClefNote extends Note {
 
     this.setRendered();
 
-    this.clef.renderText(ctx, this.getAbsoluteX(), stave.getYForLine(this.clef.line));
+    this.clef.renderText(ctx, this.getAbsoluteX() - this.x, stave.getYForLine(this.clef.line));
   }
 }
--- a/src/element.ts
+++ b/src/element.ts
@@ -75,8 +75,8 @@ export class Element {
     return Category.Element;
   }
 
-  // all Element objects keep a list of children that they are responsible and which
-  // inherit the style of their parents.
+  // Element objects keep a list of children that they are responsible for.
+  // Children inherit the style from their parents (see: setGroupStyle(s)).
   protected children: Element[] = [];
   protected static ID: number = 1000;
   protected static newID(): string {
--- a/src/element.ts
+++ b/src/element.ts
@@ -107,6 +107,8 @@ export class Element {
   protected width: number = 0;
   protected xShift: number = 0;
   protected yShift: number = 0;
+  protected x: number = 0;
+  protected y: number = 0;
 
   constructor(category?: string) {
     this.#attrs = {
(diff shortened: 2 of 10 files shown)


ronyeh:
Thanks for these. I have been looking at the other PRs but as you are aware, the bigger PRs are more difficult for a reviewer to tackle. :-) Smaller bites are easier to understand.

Apologies if the reviews are going a bit slowly. "You're coding too fast!" Hahahah.

src/element.ts
```diff
@@ -521,7 +544,11 @@ export class Element {
   renderText(ctx: RenderContext, xPos: number, yPos: number): void {
     ctx.save();
     ctx.setFont(this.textFont);
-    ctx.fillText(this.text, xPos + this.xShift, yPos + this.yShift);
+    ctx.fillText(this.text, xPos + this.x + this.xShift, yPos + this.y + this.yShift);
+    this.children.forEach((child) => {
```
ronyeh, line 548: Can we also clean up the grammar/formatting for the comment on lines 78-79 of this file? (e.g., "they are responsible for").

```js
  // all Element objects keep a list of children that they are responsible and which
  // inherit the style of their parents.
  protected children: Element[] = [];
```

Maybe something like:

```js
  // Element objects keep a list of children that they are responsible for.
  // Children inherit the style from their parents (see: setGroupStyle(s)).
  protected children: Element[] = [];
```

ronyeh (changes requested):

src/clefnote.ts
```diff
@@ -45,6 +45,6 @@ export class ClefNote extends Note {
 
     this.setRendered();
 
-    this.clef.renderText(ctx, this.getAbsoluteX(), stave.getYForLine(this.clef.line));
+    this.clef.renderText(ctx, this.getAbsoluteX() - this.x, stave.getYForLine(this.clef.line));
```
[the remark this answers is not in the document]
rvilarl: The idea is to generalize x and y in Element. I went for that when I saw noteheads (now inluded in this PR). getAbsoluteX inludes the x of the clef which is already considered in Element, I believe that Elements should also handle margins in the future.

src/element.ts
rvilarl: done

src/element.ts
```diff
@@ -521,7 +544,11 @@ export class Element {
...
+      ctx.setFont(child.textFont);
+      ctx.fillText(child.text, xPos + child.x + child.xShift, yPos + child.y + child.yShift);
```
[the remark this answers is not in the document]
rvilarl: No, they are independent. Same approach as Noteheads. They share the xPos and yPos given as parameter to RenderText.

src/element.ts
```diff
@@ -75,8 +75,8 @@ export class Element {
     return Category.Element;
   }
 
-  // all Element objects keep a list of children that they are responsible and which
-  // inherit the style of their parents.
+  // Element objects keep a list of children that they are responsible for.
+  // Children inherit the style from their parents (see: setGroupStyle(s)).protected children: Element[] = [];
```
ronyeh, line 79: There's a small copy/paste error here.

ronyeh: Remvoe the "protected children..." at the end.

rvilarl: Done.

rvilarl: Done.

ronyeh:
Thanks!

PR #68, opened 2023-08-23
notes refactoring preparation extracted from Notes (v5)

rvilarl:
No visual differences.

### DIFF
--- a/src/annotation.ts
+++ b/src/annotation.ts
@@ -68,8 +68,7 @@ export class Annotation extends Modifier {
   // Use the same padding for annotations as note head so the
   // words don't run into each other.
   static get minAnnotationPadding(): number {
-    const musicFont = Tables.currentMusicFont();
-    return musicFont.lookupMetric('noteHead.minPadding');
+    return Tables.lookupMetric('NoteHead.minPadding');
   }
   /** Arrange annotations within a `ModifierContext` */
   static format(annotations: Annotation[], state: ModifierContextState): boolean {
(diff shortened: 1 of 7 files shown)


src/stave.ts
```diff
@@ -80,14 +80,12 @@ export class Stave extends Element {
   // This is the sum of the padding that normally goes on left + right of a stave during
   // drawing. Used to size staves correctly with content width.
   static get defaultPadding(): number {
-    const musicFont = Tables.currentMusicFont();
-    return musicFont.lookupMetric('stave.padding') + musicFont.lookupMetric('stave.endPaddingMax');
+    return Tables.lookupMetric('Stave.padding') + Tables.lookupMetric('Stave.endPaddingMax');
```
ronyeh, line 83: Once migration is done, we should figure out where the `lookupMetric()` function belongs. I guess it's OK to make the `Tables` class officially public, as long as we clean it up and make the API clear & easy to use.

src/annotation.ts
```diff
@@ -68,8 +68,7 @@ export class Annotation extends Modifier {
   // Use the same padding for annotations as note head so the
   // words don't run into each other.
   static get minAnnotationPadding(): number {
-    const musicFont = Tables.currentMusicFont();
-    return musicFont.lookupMetric('noteHead.minPadding');
+    return Tables.lookupMetric('NoteHead.minPadding');
```
ronyeh, line 71: I noticed that the metric keys now have their first letter capitalized. I guess it makes the first part look more like a class name?

PR #70, opened 2023-08-25
flag (v5) extracted from note (v5)

rvilarl:
Flag and Gracenote refactored.


rvilarl:
current
pptr-Accidental Accidental_Arrangement_Special_Cases Bravura svg_current
reference
pptr-Accidental Accidental_Arrangement_Special_Cases Bravura svg_reference
current
pptr-Grace_Notes Grace_Note_Basic_with_Slurs Bravura svg_current
reference
pptr-Grace_Notes Grace_Note_Basic_with_Slurs Bravura svg_reference

ronyeh:
The current grace notes look too tiny to my eyes.... Maybe I'm getting old.

ronyeh:
From measuring the pixels, it looks like the current gracenotes are about 50-55% of the normal notes size. I measured the note heads.

In MuseScore, the gracenote is about 67-69% of the regular note's size.

I think the intended size is about 2/3, which is what the reference version looks like (I used a pixel measurement tool).

Why is the new gracenote smaller? We should strive to minimize the visual differences, since the existing V4 version of VexFlow has been "visually vetted" by many folks over the years.

ronyeh:
In this sample using Bravura SMuFL glyphs, it looks like the grace note is about 60% of the size of the normal note:
image



The code points are from:
- https://w3c.github.io/smufl/latest/tables/common-ornaments.html?highlight=grace#common-ornaments-ue560ue56f
- https://w3c.github.io/smufl/latest/tables/individual-notes.html

src/gracenote.ts
```diff
@@ -46,32 +46,22 @@ export class GraceNote extends StaveNote {
...
-
-    return 0;
+    let ret = super.getStemExtension();
+    ret = Stem.HEIGHT * this.getStaveNoteScale() - Stem.HEIGHT + ret;
+    return ret;
   }
 
   getStaveNoteScale(): number {
-    return this.renderOptions.glyphFontScale / Tables.NOTATION_FONT_SCALE;
+    return 2 / 3;
```
ronyeh, line 55: I think 2/3 should be the correct target scale for grace notes. However, the final render seems smaller than 2/3.

src/gracenote.ts
```diff
@@ -80,56 +70,40 @@ export class GraceNote extends StaveNote {
...
+            x2: x + noteHeadWidth,
+            y2: y + noteHeadWidth,
+          };
+        else
+          slashBBox = {
+            x1: x - noteHeadWidth,
+            y1: y + noteHeadWidth,
+            x2: x + noteHeadWidth,
+            y2: y - noteHeadWidth,
+          };
```
ronyeh, line 99: This is slightly confusing when skimming the code. Please use extra braces { } for the if/else blocks.

ronyeh: My two reasons for preferring braces:
- In the future if someone needs to add a line of code inside the if/else case, it is easier to do so.
- It is easier to read the code and visually inspect for logic errors.

src/stavenote.ts
```diff
@@ -1120,28 +1120,22 @@ export class StaveNote extends StemmableNote {
...
       // Alternatively, we could shorten the stem.
       const flagY =
         this.getStemDirection() === Stem.DOWN
           ? // Down stems are below the note head and have flags on the right.
-            yTop -
-            noteStemHeight +
-            2 -
-            (this.glyphProps ? this.glyphProps.stemDownExtension : 0) * this.getStaveNoteScale() -
-            (this.flag?.getMetrics().yShift ?? 0) * (1 - this.getStaveNoteScale())
+            yTop - noteStemHeight - this.flag.getTextMetrics().actualBoundingBoxDescent
```
ronyeh, line 1131: So these two lines of code (1131 and 1133) removed the magic constant of 2. Does that mean we no longer need it to fix the stem issue?

If so, you can remove the block comment above ("What's with the magic +/- 2.....").

src/stemmablenote.ts
```diff
@@ -189,14 +202,12 @@ export abstract class StemmableNote extends Note {
...
-      return glyphProps.stemBeamExtension;
-    }
-
-    if (glyphProps) {
-      return this.getStemDirection() === Stem.UP ? glyphProps.stemUpExtension : glyphProps.stemDownExtension;
+      return glyphProps.stemBeamExtension * this.getStaveNoteScale();
     }
 
-    return 0;
+    return this.flag.getHeight() > Stem.HEIGHT * this.getStaveNoteScale()
```
ronyeh, line 208: Can we add a comment here? I don't understand this line.

Maybe consider extracting two local vars for:
`this.flag.getHeight()` and 
`Stem.HEIGHT * this.getStaveNoteScale()`

ronyeh (changes requested):

ronyeh:
Review done, with my comments above. Primarily I'd like to get the grace notes looking like the used to. Right now they are too small.

rvilarl:
> Review done, with my comments above. Primarily I'd like to get the grace notes looking like the used to. Right now they are too small.

The size difference has been introduced with the extraction from note (v5). There the proportion is right. I will look into that.

src/stemmablenote.ts
```diff
@@ -51,13 +53,19 @@ export abstract class StemmableNote extends Note {
...
-      const flagCode = this.getStemDirection() === Stem.DOWN ? glyphProps.codeFlagDownstem : glyphProps.codeFlagUpstem;
-
-      if (flagCode) this.flag = new Glyph(flagCode, this.renderOptions.glyphFontScale, { category });
+      const flagCode =
+        this.getStemDirection() === Stem.DOWN ? glyphProps.codeFlagDownstem ?? '' : glyphProps.codeFlagUpstem ?? '';
+
+      this.flag.setText(flagCode);
+      // #FIXME: HACK to use 30 as fontSize default rather than 39 used with glyphs.
+      // HACK-BEGIN
+      this.flag.fontSize = (this.renderOptions.glyphFontScale / 39) * 30;
```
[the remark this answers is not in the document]
rvilarl: 30 is the value that produces the glyphs at the write size for our staves. I made the investigation at the beginning of the migration with the Clefs.

src/stavenote.ts
rvilarl: done

src/stemmablenote.ts
rvilarl: done

src/gracenote.ts
rvilarl: done

src/gracenote.ts
rvilarl: Hi Ron, please ignore this one. As I said below, this is corrected once vf5/note is fully integarted  and I do not feel like looking for a hack for the intermediate step. I will set an issue.

PR #80, opened 2023-09-02
operators fix

rvilarl:
No visual differences.

### DIFF
--- a/src/stavehairpin.ts
+++ b/src/stavehairpin.ts
@@ -67,7 +67,7 @@ export class StaveHairpin extends Element {
   ): void {
     const ppt = formatter.pixelsPerTick;
 
-    if (ppt == null) {
+    if (ppt === null) {
       throw new RuntimeError('BadArguments', 'A valid Formatter must be provide to draw offsets by ticks.');
     }
 
@@ -132,14 +132,7 @@ export class StaveHairpin extends Element {
   }
 
   setRenderOptions(options: StaveHairpinRenderOptions): this {
-    if (
-      options.height != null &&
-      options.yShift != null &&
-      options.leftShiftPx != null &&
-      options.rightShiftPx != null
-    ) {
-      this.renderOptions = options;
-    }
+    this.renderOptions = options;
     return this;
   }
 
--- a/src/stavenote.ts
+++ b/src/stavenote.ts
@@ -152,7 +152,8 @@ export class StaveNote extends StemmableNote {
     const draw = [false, false, false];
 
     for (let i = 0; i < notesList.length; i++) {
-      draw[i] = notesList[i].note.renderOptions.draw === false ? false : true;
+      // If .draw is true or undefined, we set draw[i] = true
+      draw[i] = notesList[i].note.renderOptions.draw !== false;
     }
 
     if (draw[0] && draw[1] && draw[2]) {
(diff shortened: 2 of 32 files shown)


ronyeh:
Is this independent from #35? If so, I can review/merge this first.

rvilarl:
yes, it is independent from #35 it is related to your comments on operators

src/stavehairpin.ts
```diff
@@ -132,14 +132,7 @@ export class StaveHairpin extends Element {
...
   setRenderOptions(options: StaveHairpinRenderOptions): this {
-    if (
-      options.height != null &&
-      options.yShift != null &&
-      options.leftShiftPx != null &&
-      options.rightShiftPx != null
-    ) {
-      this.renderOptions = options;
-    }
+    this.renderOptions = options;
```
ronyeh, line 135: Just checking on the reasoning behind this deletion.

So now it's OK to `null` out certain options?

rvilarl: According to the type definition, these elements cannot be null.

src/stavenote.ts
```diff
@@ -152,7 +152,7 @@ export class StaveNote extends StemmableNote {
     const draw = [false, false, false];
 
     for (let i = 0; i < notesList.length; i++) {
-      draw[i] = notesList[i].note.renderOptions.draw == false ? false : true;
+      draw[i] = notesList[i].note.renderOptions.draw === false ? false : true;
```
ronyeh, line 155: This is a totally weird line of code :-). I guess it is because .draw can be true or false or undefined.

```
  public renderOptions: {
    drawStemThroughStave?: boolean;
    draw?: boolean;
```

Can we rewrite it as:

```ts
// If .draw is true or undefined, we set draw[i] = true
draw[i] = notesList[i].note.renderOptions.draw !== false;
```

src/stemmablenote.ts
```diff
@@ -60,7 +60,7 @@ export abstract class StemmableNote extends Note {
         // codeFlagDown = codeFlagUp + 1,, if not defined, code should be 0
         this.getStemDirection() === Stem.DOWN
           ? String.fromCodePoint((glyphProps.codeFlagUp?.codePointAt(0) ?? -1) + 1)
-          : glyphProps.codeFlagUp ?? '\u0000';
+          : glyphProps.codeFlagUp  || '\u0000';
```
ronyeh, line 63: How come this was switched the other way? If so, please add a comment to explain the reasoning, as it is not obvious.

The comment has a small typo. Please change to:
```
// codeFlagDown = codeFlagUp + 1
// if codeFlagUp is undefined, codePoint will be 0
```

src/stemmablenote.ts
```diff
@@ -100,19 +100,19 @@ export abstract class StemmableNote extends Note {
     // if note is flagged, cannot shorten beam
     switch (this.duration) {
       case '8':
-        if (this.beam == undefined) length = 35;
+        if (this.beam === undefined) length = 35;
```
ronyeh, line 103: Can we extract this:

```js
const beamIsUndefined = this.beam === undefined;
```

Also, why does case '8': look different from all the other cases?

ronyeh (changes requested):
Looks good. Minor changes / comments. Thanks for the work!

src/stavenote.ts
rvilarl: done

src/stemmablenote.ts
rvilarl: reverted, comment updated.

src/stemmablenote.ts
rvilarl: refactored, no visual differences

ronyeh (approved):

PR #81, opened 2023-09-02
staverepetition (v5)

rvilarl:
Simpler but with differences

### DIFF
--- a/src/staverepetition.ts
+++ b/src/staverepetition.ts
@@ -62,10 +62,10 @@ export class Repetition extends StaveModifier {
         this.drawSymbolText(stave, x, 'Coda', true);
         break;
       case Repetition.type.SEGNO_LEFT:
-        this.drawSignoFixed(stave, x);
+        this.drawSegnoFixed(stave, x);
         break;
       case Repetition.type.SEGNO_RIGHT:
-        this.drawSignoFixed(stave, x + stave.getWidth());
+        this.drawSegnoFixed(stave, x + stave.getWidth());
         break;
       case Repetition.type.DC:
         this.drawSymbolText(stave, x, 'D.C.', false);
@@ -105,7 +105,7 @@ export class Repetition extends StaveModifier {
     return this;
   }
 
-  drawSignoFixed(stave: Stave, x: number): this {
+  drawSegnoFixed(stave: Stave, x: number): this {
     const y = stave.getYForTopText(stave.getNumLines());
     this.text = '\ue047' /*segno*/;
     this.renderText(stave.checkContext(), x, y + Tables.lookupMetric('Repetition.segno.offsetY'));
--- a/src/staverepetition.ts
+++ b/src/staverepetition.ts
@@ -1,8 +1,6 @@
 // Copyright (c) 2023-present VexFlow contributors: https://github.com/vexflow/vexflow/graphs/contributors
 // @author: Larry Kuhns 2011
 
-import { Font } from './font';
-import { Glyph } from './glyph';
 import { Stave } from './stave';
 import { StaveModifier } from './stavemodifier';
 import { Tables } from './tables';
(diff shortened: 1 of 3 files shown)


rvilarl:
current
Stave Volta___Modifier_Measure_Test Bravura svg_current
reference
Stave Volta___Modifier_Measure_Test Bravura svg_reference

current
Stave Multiple_Staves_Volta_Test Bravura svg_current
reference
Stave Multiple_Staves_Volta_Test Bravura svg_reference

rvilarl:
Gonville glyphs are too small.
Bravura
Stave Multiple_Staves_Volta_Test Bravura svg_current
Gonville
Stave Multiple_Staves_Volta_Test Gonville svg_current
Leland
Stave Multiple_Staves_Volta_Test Leland svg_current
Musejazz
Stave Multiple_Staves_Volta_Test MuseJazz svg_current
Petaluma
Stave Multiple_Staves_Volta_Test Petaluma svg_current

src/staverepetition.ts
```diff
@@ -101,95 +99,52 @@ export class Repetition extends StaveModifier {
...
-      'coda',
-      { category: 'coda' }
-    );
+    const y = stave.getYForTopText(stave.getNumLines());
+    this.text = '\ue048' /*coda*/;
+    this.renderText(stave.checkContext(), x, y + Tables.lookupMetric('Repetition.coda.offsetY'));
     return this;
   }
 
   drawSignoFixed(stave: Stave, x: number): this {
```
ronyeh, line 108: Since we're here, can we rename this method? It misspells it as signo.

drawSignoFixed => drawSegnoFixed

ronyeh: This typo was in version 3.0.9, so we should document this breaking change (along with all our other breaking changes). :-)

https://github.com/0xfe/vexflow/blob/3.0.9/src/staverepetition.js#L95

ronyeh: OK, I created a wiki page to document breaking changes :-)

https://github.com/vexflow/vexflow/wiki/Breaking-Changes

ronyeh (changes requested):
Looks fine. Small changes requested. I'm also awaiting your Gonville / Gootville investigation and then I can merge this.

src/staverepetition.ts
rvilarl: Done. 
Is this really a breaking change? The method is only used internally. We have not started protecting internal member functions. Should we?

rvilarl:
> Looks fine. Small changes requested. I'm also awaiting your Gonville / Gootville investigation and then I can merge this.

Gootville is equaly tiny.

src/staverepetition.ts
ronyeh: For more flexibility, I prefer not protecting internal member functions unless there is a good reason for it to be private.

The reason is that a developer can subclass / override methods if we don't protect them. Or they can do something cool / unexpected by calling our internal methods. That makes the library more useful in general. 

But if we make every method private, then developers are limited in what they can do (unless they download from the main branch and edit the code themselves).

ronyeh: Also, even if it is not a "breaking change" by strict definition, I figure it doesn't hurt to document this somewhere, in case someone else asks later on.

src/staverepetition.ts
```diff
@@ -101,95 +99,52 @@ export class Repetition extends StaveModifier {
...
-    ctx.save();
-    ctx.setFont(this.textFont);
-
     let textX = 0;
-    let symbolX = 0;
-    const modifierWidth = stave.getNoteStartX() - this.x;
+
+    this.text = text;
+    if (drawCoda) {
+      this.text += ' \ue048' /*coda*/;
```
ronyeh, line 121: Is this leading space intentional? It looks unexpected. If it is intentional, then please document the reason why with a comment here.

ronyeh: Is this because this method is for drawing D.S. al ⊕ so you need the space before the coda string? 

Can you add that as a comment here? I'll go ahead and merge, so you can add it to your next PR.

src/staverepetition.ts
rvilarl: @ronyeh I meant to make methods protected, not private. Protected methods can be overriden.

PR #82, opened 2023-09-02
new fonts

rvilarl:
I think that it is the right time to add new fonts to check if v5 is working well. The tests take double the time but I still think it is worthy to see if we are including generic problems. I added:
- Gootville
- Finale Ash
- Finale Broadway
- Finale Maestro

### DIFF
--- a/src/fonts/load_finaleash.ts
+++ b/src/fonts/load_finaleash.ts
@@ -1,0 +1,20 @@
+// Copyright (c) 2023-present VexFlow contributors: https://github.com/vexflow/vexflow/graphs/contributors
+// MIT License
+
+// If you are adding a new music engraving font, search for instances of ADD_MUSIC_FONT throughout the code base.
+// To compile your new font into vexflow.js, take a look at src/fonts/load_all.ts
+// You can export a font module which can be dynamically loaded by vexflow-core.js (see: Gruntfile.js).
+
+// ADD_MUSIC_FONT
+// To add a new music engraving font XXX, make a copy of this file and name it load_xxx.ts.
+// Then you will need to generate xxx_glyphs.ts and xxx_metrics.ts.
+// xxx_glyphs.ts is created by tools/fonts/fontgen_smufl.js
+// xxx_metrics.ts is created by hand. You could copy bravura_metrics.ts and modify/remove/add entries where necessary.
+
+import { Font } from '../font';
+import { BravuraFont } from './bravura_glyphs';
+import { CommonMetrics } from './common_metrics';
+
+export function loadFinaleAsh() {
+  Font.load('Finale Ash', BravuraFont, CommonMetrics);
+}
--- a/tests/vexflow_test_helpers.ts
+++ b/tests/vexflow_test_helpers.ts
@@ -117,9 +117,9 @@ const SVG_TEST_CONFIG = {
   testType: 'SVG',
   fontStacks: [
     'Bravura',
-    'FinaleAsh',
-    'FinaleBroadway',
-    'FinaleMaestro',
+    'Finale Ash',
+    'Finale Broadway',
+    'Finale Maestro',
     'Gonville',
     'Gootville',
     'Leland',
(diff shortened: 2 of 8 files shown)


rvilarl:
Bravura
Formatter Proportional_Formatting___No_Tuning Bravura svg
Finale Ash
Formatter Proportional_Formatting___No_Tuning FinaleAsh svg
Finale Broadway
Formatter Proportional_Formatting___No_Tuning FinaleBroadway svg
Finale Maestro
Formatter Proportional_Formatting___No_Tuning FinaleMaestro svg
Gonville
Formatter Proportional_Formatting___No_Tuning Gonville svg
Gootville
Formatter Proportional_Formatting___No_Tuning Gootville svg
Leland
Formatter Proportional_Formatting___No_Tuning Leland svg
Musejazz
Formatter Proportional_Formatting___No_Tuning MuseJazz svg
Petaluma
Formatter Proportional_Formatting___No_Tuning Petaluma svg

rvilarl:
@sschmidtu @jaredjj3 @0xfe @AaronDavidNewman @mscuthbert @ronyeh do you like how it is looking?

0xfe:
Fonts look great. Those time signatures in Muzejazz and Petaluma seem off though.

rvilarl:
> Fonts look great. Those time signatures in Muzejazz and Petaluma seem off though.

Petaluma was discussed with the authors and it is supposed to be like that. However I agree Musejazz and also Finale Boradway are off. Thanks for the feedback!

src/fonts/load_finaleash.ts
```diff
@@ -0,0 +1,20 @@
...
+// Then you will need to generate xxx_glyphs.ts and xxx_metrics.ts.
+// xxx_glyphs.ts is created by tools/fonts/fontgen_smufl.js
+// xxx_metrics.ts is created by hand. You could copy bravura_metrics.ts and modify/remove/add entries where necessary.
+
+import { Font } from '../font';
+import { BravuraFont } from './bravura_glyphs';
+import { CommonMetrics } from './common_metrics';
+
+export function loadFinaleAsh() {
+  Font.load('Finale Ash', BravuraFont, CommonMetrics);
```
ronyeh, line 19: I assume eventually, we won't need to load the glyphs from BravuraFont anymore?

tests/vexflow_test_helpers.ts
```diff
@@ -200,6 +210,10 @@ export class VexFlowTests {
     Petaluma: ['Petaluma', 'Gonville', 'Bravura', 'Custom'],
     Leland: ['Leland', 'Bravura', 'Custom'],
     MuseJazz: ['MuseJazz'],
```
ronyeh, line 212: During testing / migration, we should remove the fallback fonts in the font stacks.

```ts
    Bravura: ['Bravura'/*, 'Custom'*/],
    Gonville: ['Gonville'/*, 'Bravura', 'Custom'*/],
    Petaluma: ['Petaluma'/*, 'Gonville', 'Bravura', 'Custom'*/],
    Leland: ['Leland'/*, 'Bravura', 'Custom'*/],
```

That way, we can easily see which glyphs are missing from each font.

tests/vexflow_test_helpers.ts
```diff
@@ -115,7 +115,17 @@ const SVG_TEST_CONFIG = {
   backend: Renderer.Backends.SVG,
   tagName: 'div',
   testType: 'SVG',
-  fontStacks: ['Bravura', 'Gonville', 'Petaluma', 'Leland', 'MuseJazz'],
+  fontStacks: [
+    'Bravura',
+    'FinaleAsh',
+    'FinaleBroadway',
+    'FinaleMaestro',
```
ronyeh, line 122: If the official font names have a space, why don't we be consistent and keep the space? Otherwise it'll get confusing later (remember with Roboto Slab and Petaluma Script)?

src/fonts/load_finaleash.ts
rvilarl: This is right, all the loadXXX functions will go away.

tests/vexflow_test_helpers.ts
rvilarl: We can do that once Glyph goes away. Otherwise we will get a RunTime error.

tests/vexflow_test_helpers.ts
rvilarl: Done

ronyeh (approved):

PR #84, opened 2023-09-02
parenthesis (v5)

rvilarl:
Minimal visual differences:

current
EasyScore Draw_Parenthesised Bravura svg_current
reference
EasyScore Draw_Parenthesised Bravura svg_reference

### DIFF
--- a/src/parenthesis.ts
+++ b/src/parenthesis.ts
@@ -2,11 +2,9 @@
 // @author Rodrigo Vilar
 // MIT License
 
-import { Glyph } from './glyph';
 import { Modifier, ModifierPosition } from './modifier';
 import { ModifierContextState } from './modifiercontext';
 import { Note } from './note';
-import { Tables } from './tables';
 import { Category, isGraceNote } from './typeguard';
 
 /** Parenthesis implements parenthesis modifiers for notes. */
@@ -14,8 +12,6 @@ export class Parenthesis extends Modifier {
   static get CATEGORY(): string {
     return Category.Parenthesis;
   }
-
-  protected point: number;
 
   /** Add parentheses to the notes. */
   static buildAndAttach(notes: Note[]): void {
@@ -47,7 +43,7 @@ export class Parenthesis extends Modifier {
         xWidthR = xWidthR > shift + parenthesis.width ? xWidthR : shift + parenthesis.width;
       }
       if (pos === ModifierPosition.LEFT) {
-        shift = note.getLeftParenthesisPx(index);
+        shift = note.getLeftParenthesisPx(index) + parenthesis.width;
         xWidthL = xWidthL > shift + parenthesis.width ? xWidthL : shift + parenthesis.width;
       }
       parenthesis.setXShift(shift);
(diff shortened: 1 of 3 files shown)


src/parenthesis.ts
```diff
@@ -67,19 +63,22 @@ export class Parenthesis extends Modifier {
     super();
 
     this.position = position ?? Modifier.Position.LEFT;
+    if (this.position == Modifier.Position.RIGHT) {
+      this.text = '\uE0F6'; /*noteheadParenthesisRight*/
+    } else if (this.position == Modifier.Position.LEFT) {
+      this.text = '\uE0F5'; /*noteheadParenthesisLeft*/
+    }
```
ronyeh, line 70: Let's use === by default.

src/parenthesis.ts
```diff
@@ -67,19 +63,22 @@ export class Parenthesis extends Modifier {
...
 
-    this.point = Tables.currentMusicFont().lookupMetric('parenthesis.default.point') ?? Note.getPoint('default');
-    this.setWidth(Tables.currentMusicFont().lookupMetric('parenthesis.default.width'));
+    this.setFontSize(Note.getPoint('default'));
```
ronyeh, line 72: This can be a different PR, but I think Note.getPoint() could have a clearer method name. Does it relate to font size, or to a point in 2D space?

src/parenthesis.ts
```diff
@@ -67,19 +63,22 @@ export class Parenthesis extends Modifier {
...
   setNote(note: Note): this {
     this.note = note;
-    this.point = Tables.currentMusicFont().lookupMetric('parenthesis.default.point') ?? Note.getPoint('default');
-    this.setWidth(Tables.currentMusicFont().lookupMetric('parenthesis.default.width'));
     if (isGraceNote(note)) {
-      this.point = Tables.currentMusicFont().lookupMetric('parenthesis.gracenote.point') ?? Note.getPoint('gracenote');
-      this.setWidth(Tables.currentMusicFont().lookupMetric('parenthesis.gracenote.width'));
+      this.setFontSize(Note.getPoint('gracenote'));
+    } else {
+      this.setFontSize(Note.getPoint('default'));
```
ronyeh, line 81: Why isn't it named Note.getFontSize()?

ronyeh (changes requested):
Just some questions, but otherwise looks OK.

src/parenthesis.ts
rvilarl: :) done

src/parenthesis.ts
rvilarl: Very good idea! rather get the font from the note: note.getFont()

src/parenthesis.ts
rvilarl: No longer used.

src/parenthesis.ts
rvilarl: And remove note.getPoint

ronyeh:
Thanks!

I'll get back to reviewing sometime in the middle of next week. Have a nice weekend (and take a deserved break, haha)!

PR #85, opened 2023-09-02
remove glyph in tests

rvilarl:
Visual changes due to the different measure.

current
KeySignature Cancelled_key__for_each_clef__test Bravura svg_current
reference
KeySignature Cancelled_key__for_each_clef__test Bravura svg_reference
current
Ornament Jazz_Ornaments Bravura svg_current
reference
Ornament Jazz_Ornaments Bravura svg_reference

### DIFF
--- a/tests/key_clef_tests.ts
+++ b/tests/key_clef_tests.ts
@@ -6,7 +6,7 @@
 
 import { MAJOR_KEYS, MINOR_KEYS, TestOptions, VexFlowTests } from './vexflow_test_helpers';
 
-import { Glyph } from '../src/glyph';
+import { Element } from '../src/element';
 import { KeySignature } from '../src/keysignature';
 import { ContextBuilder } from '../src/renderer';
 import { Stave } from '../src/stave';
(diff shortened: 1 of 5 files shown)


tests/stringnumber_tests.ts
```diff
@@ -275,8 +275,10 @@ function multi(options: TestOptions): void {
 
 function drawAccidentals(options: TestOptions): void {
   const f = VexFlowTests.makeFactory(options, 750);
-  const glyphScale = 39; // default font scale
-  const clefWidth = Glyph.getWidth('gClef', glyphScale); // widest clef
+  const el = new Element();
+  el.setText(String.fromCharCode(0xe050));
```
ronyeh, line 279: Can we just use the unicode string directly instead of converting from char code? Also, we should include the SMuFL glyph name for future reference. Same comment goes for the above, in ornament_tests.ts.

tests/keysignature_tests.ts
```diff
@@ -29,12 +29,18 @@ const KeySignatureTests = {
   },
 };
 
+function getWidth(code: number) {
+  const el = new Element();
+  el.setText(String.fromCharCode(code));
```
ronyeh, line 34: I think in our examples, we should always use String.fromCodePoint() unless there is a good performance related reason to use fromCharCode(). The newer method handles all unicode code points properly.

tests/ornament_tests.ts
```diff
@@ -237,7 +237,10 @@ function drawOrnamentsWithAccidentals(options: TestOptions): void {
 }
 
 function jazzOrnaments(options: TestOptions): void {
-  const clefWidth = Glyph.getWidth('gClef', 38); // widest clef
+  const el = new Element();
+  el.setText(String.fromCharCode(0xe050));
```
ronyeh, line 241: We should prefer fromCodePoint()

tests/key_clef_tests.ts
```diff
@@ -22,13 +22,19 @@ const ClefKeySignatureTests = {
   },
 };
 
+function getWidth(code: number) {
+  const el = new Element();
+  el.setText(String.fromCharCode(code));
```
ronyeh, line 27: String.fromCodePoint()

tests/keysignature_tests.ts
rvilarl: Agree but changed to unicode value, which should be the preferred method.

tests/ornament_tests.ts
rvilarl: changed to unicode.

tests/key_clef_tests.ts
rvilarl: Changed to unicode

tests/stringnumber_tests.ts
ronyeh, line 279: @rvilarl  Why is this one still using the numeric code point? I'll merge, but can you fix this in the next PR?

ronyeh (changes requested):
One suggested change, that can be included in the next PR.

tests/stringnumber_tests.ts
ronyeh: We should look for uses of fromCharCode() and fromCodePoint() and make sure they're really needed.

PR #87, opened 2023-09-03
remove textformatter

rvilarl:
TextFormatter no longer required. No visual differences.


rvilarl:
@AaronDavidNewman is there any API missing with this change that I should add to Element?

src/textformatter.ts
```diff
@@ -1,346 +0,0 @@
...
-      extent.yMax = Math.max(extent.yMax, curY.yMax);
-      extent.height = extent.yMax - extent.yMin;
-    }
-    entry[str] = extent;
-    return extent;
-  }
-  /**
-   * Retrieve the total width of `text` in `em` units.
-   */
-  getWidthForTextInEm(text: string): number {
```
ronyeh, line 303: Is your measureText() related to this?

rvilarl: Is this question for me?

ronyeh: Yes, I was just thinking about whether these methods have any uses. Your new v5 approach relies heavily on a measureText() method. I wonder if others also have use for methods that are similar (for layout / UI purposes).

This method is called getWidthForTextInEm() so it sounds like it might be a useful utility method.

ronyeh: In any case, if you think this class has no uses for now, we can delete it. But if anyone complains, we can always restore it (from the most recent version, or from the v4 repository).

ronyeh:
@AaronDavidNewman I'm just checking in with you to make sure we are not removing something important here. Are you able to do what you need to do with Smoosic with the current v5 APIs, if we remove these files?

src/textformatter.ts
rvilarl: In VexFlow v4 Glyph focuses on SMuFL metrics and TextFormatter on other Fonts (i.e.: Roboto Slab') In VexFlow v5 we considered both glyphs and standard characters text and it is all handled by Element.

AaronDavidNewman:
I am not able to keep up with the changes.  I think I had a plan for this but it will be probably next week before I can make the updates to Smoosic.  Don't wait on my account.

ronyeh:
OK, I'm going to merge this but if you run into a huge issue with smoosic, just let us know and we can revert / reintroduce any APIs you need (assuming there isn't a good alternative).

PR #88, opened 2023-09-03
staveconnector (v5)

rvilarl:
StaveConnector & StaveText ported. Visual changes:

current
StaveConnector Combined_Draw_Test Petaluma svg_current
reference
StaveConnector Combined_Draw_Test Petaluma svg_reference

### DIFF
--- a/src/staveconnector.ts
+++ b/src/staveconnector.ts
@@ -2,7 +2,6 @@
 // MIT License
 
 import { Element } from './element';
-import { Glyph } from './glyph';
 import { RenderContext } from './rendercontext';
 import { Stave } from './stave';
 import { Tables } from './tables';
@@ -103,11 +102,7 @@ export class StaveConnector extends Element {
     none: StaveConnector.type.NONE,
   } as const;
 
-  protected width: number;
-  protected texts: {
-    content: string;
-    options: { shiftX: number; shiftY: number };
-  }[];
+  protected texts: Element[];
 
   protected type: (typeof StaveConnector)['type'][keyof (typeof StaveConnector)['type']];
 
@@ -115,13 +110,10 @@ export class StaveConnector extends Element {
   readonly bottomStave: Stave;
   readonly thickness: number;
 
-  protected xShift: number;
-
   constructor(topStave: Stave, bottomStave: Stave) {
     super();
 
     this.thickness = Tables.STAVE_LINE_THICKNESS;
-    this.width = 3;
     this.topStave = topStave;
     this.bottomStave = bottomStave;
     this.type = StaveConnector.type.DOUBLE;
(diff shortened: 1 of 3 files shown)


src/tables.ts
```diff
@@ -79,8 +79,9 @@ export const CommonMetrics: Record<string, any> = {
   },
 
   StaveConnector: {
-    fontFamily: 'Times New Roman, serif',
-    fontSize: 16,
+    text: {
+      fontSize: 16,
+    },
```
ronyeh, line 84: I don't understand why this fontSize is under a text { } sub element, while below on line 109, the fontSize is declared directly on the StaveText element.

src/stavetext.ts
```diff
@@ -29,79 +21,50 @@ export class StaveText extends StaveModifier {
...
+          y = stave.getYForBottomText(2);
         }
         break;
       default:
         throw new RuntimeError('InvalidPosition', 'Value Must be in Modifier.Position.');
     }
 
-    ctx.fillText('' + this.text, x, y + 4);
-    ctx.restore();
+    this.renderText(ctx, x, y + 4);
```
ronyeh, line 67: Do we know why this magic number +4 is here? If so, can we add a comment to explain it? :-)

src/staveconnector.ts
```diff
@@ -255,9 +233,11 @@ export class StaveConnector extends Element {
...
         botY += 6;
         attachmentHeight = botY - topY;
-        Glyph.renderGlyph(ctx, topX - 5, topY, 40, 'bracketTop');
-        Glyph.renderGlyph(ctx, topX - 5, botY, 40, 'bracketBottom');
-        topX -= this.width + 2;
+        element.setText('\ue003' /* bracketTop */);
+        element.renderText(ctx, topX - 5, topY);
+        element.setText('\ue004' /* bracketBottom */);
+        element.renderText(ctx, topX - 5, botY);
+        topX -= 5;
```
ronyeh, line 240: Can we explain (via comments) what these magic numbers -5 do? Also, why do we need to modify topX to shift it to the left by 5?

ronyeh (changes requested):
Mostly looks good. I like how everything is now just Element (or collections of children Elements).

src/tables.ts
rvilarl: StaveText is text only, as a consequence, no distinction is required. StaveConnector consist oh glyphs and text.

src/stavetext.ts
rvilarl: I do not know. I did not look into that.

src/staveconnector.ts
rvilarl: I do not know. I did not look into that. There was two magic numbers 3 and 2 and I just replaced by 5.

src/stavetext.ts
ronyeh: OK, later on we should play around with the magic numbers in the codebase to see what they all do, hahah. We can add comments or extract them into the Metrics file if they are something that should be customizable.

PR #89, opened 2023-09-03
stroke (v5)

rvilarl:
Stroke (v5). This was more complex because it required rotation of glyphs.

### DIFF
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -43,6 +43,7 @@ export class CanvasContext extends RenderContext {
   /**  The 2D rendering context from the Canvas API. Forward method calls to this object. */
   context2D: CanvasRenderingContext2D | OffscreenCanvasRenderingContext2D;
 
+  curTransfrom: DOMMatrix;
   /**
    * The HTMLCanvasElement or OffscreenCanvas that is associated with the above context.
    * If there was no associated `<canvas>` element, just store the default WIDTH / HEIGHT.
@@ -85,6 +86,7 @@ export class CanvasContext extends RenderContext {
   constructor(context: CanvasRenderingContext2D | OffscreenCanvasRenderingContext2D) {
     super();
     this.context2D = context;
+    this.curTransfrom = context.getTransform();
     if (!context.canvas) {
       this.canvas = {
         width: CanvasContext.WIDTH,
@@ -109,6 +111,17 @@ export class CanvasContext extends RenderContext {
 
   closeGroup(): void {
     // Containers not implemented.
+  }
+
+  openRotation(angle: number, x: number, y: number) {
+    this.curTransfrom = this.context2D.getTransform();
+    this.context2D.translate(x,y);
+    this.context2D.rotate((angle * Math.PI) / 180);
+    this.context2D.translate(-x, -y);
+  }
+
+  closeRotation() {
+    this.context2D.setTransform(this.curTransfrom);
   }
 
   // eslint-disable-next-line
--- a/src/canvascontext.ts
+++ b/src/canvascontext.ts
@@ -113,10 +113,10 @@ export class CanvasContext extends RenderContext {
     // Containers not implemented.
   }
 
-  openRotation(angle: number, x: number, y: number) {
+  openRotation(angleDegrees: number, x: number, y: number) {
     this.curTransfrom = this.context2D.getTransform();
     this.context2D.translate(x,y);
-    this.context2D.rotate((angle * Math.PI) / 180);
+    this.context2D.rotate((angleDegrees * Math.PI) / 180);
     this.context2D.translate(-x, -y);
   }
 
--- a/src/strokes.ts
+++ b/src/strokes.ts
@@ -161,12 +161,13 @@ export class Stroke extends Modifier {
     if (this.type === Stroke.Type.BRUSH_DOWN || this.type === Stroke.Type.BRUSH_UP) {
       ctx.fillRect(x + this.xShift, topY, 1, botY - topY);
     } else {
+      // Select the wiggle glyph depending on the arrow direction
       const lineGlyph =
-        arrow === '\ueb7c' /*arrowheadBlackDown*/ ? '\ueaa9' /*wiggleArpeggiatoUp*/ : '\ueaaa'; /*wiggleArpeggiatoDown*/
+        arrow === '\ueb7c' /*arrowheadBlackDown*/ ? '\ueaaa' /*wiggleArpeggiatoDown*/ : '\ueaa9'; /*wiggleArpeggiatoUp*/
       let txt = '';
       const el = new Element();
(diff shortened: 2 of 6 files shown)


rvilarl:
current
Strokes Strokes___Brush_Roll_Rasquedo Bravura svg_current
reference
Strokes Strokes___Brush_Roll_Rasquedo Bravura svg_reference

ronyeh:
Interesting! Can you point out which glyphs are rotated?

ronyeh:
The arrow heads are better aligned in the old reference version.

In the current version, the up arrow heads look like they need to be offset vertically a little (so that they sit centered within the wavy part of the squiggle).

image


The down arrow looks almost centered, but the up arrowhead is offset to the left.

rvilarl:
> Interesting! Can you point out which glyphs are rotated?

the wavy part of the squiggles.  These wavy glyphs have no vertical version in SMuFL. Now it would also be possible to use a lot of different types with a minor modification to this class. See https://w3c.github.io/smufl/latest/tables/multi-segment-lines.html

src/strokes.ts
```diff
@@ -130,114 +130,87 @@ export class Stroke extends Modifier {
...
+      let txt = '';
+      const el = new Element();
+      // 2 pixels overlap of arpeggiato glyphs
+      for (; el.getWidth() < botY - topY; ) {
+        txt += lineGlyph;
+        el.setText(txt);
+        el.measureText();
+      }
+      if (
+        this.type === Stroke.Type.RASQUEDO_DOWN ||
```
ronyeh, line 175: This has probably been in the code base for a while, but I think Rasgueado is misspelled. :-)

Our typo doesn't match any of the alternative spelling on the wikipedia page: https://en.wikipedia.org/wiki/Rasgueado

The closest one is Rasqueado but we are still missing a letter a. I would suggest a separate PR to fix this (to use the official wikipedia spelling).

ronyeh: It has been around since 3.0.9, so we could (maybe) consider supporting the typo in addition to the more correct spelling. I'm actually OK with just breaking backwards compatibility, since lots of breaking changes are happening anyways. We just need to document it.

https://github.com/0xfe/vexflow/blob/00ec15c67ff333ea49f4d3defbd9e22374c03684/src/strokes.js#L21-L22

src/canvascontext.ts
```diff
@@ -111,6 +113,17 @@ export class CanvasContext extends RenderContext {
     // Containers not implemented.
   }
 
+  openRotation(angle: number, x: number, y: number) {
```
ronyeh, line 116: Maybe we should rename angle to `angleDegrees` to be more explicit.

ronyeh: Or should we instead accept radians? Then calls to openRotation() below can pass in `Math.PI * 0.5`...?

src/strokes.ts
```diff
@@ -130,114 +130,87 @@ export class Stroke extends Modifier {
...
     if (this.type === Stroke.Type.BRUSH_DOWN || this.type === Stroke.Type.BRUSH_UP) {
       ctx.fillRect(x + this.xShift, topY, 1, botY - topY);
     } else {
-      type = 'Wiggly';
-      if (isStaveNote(note)) {
-        for (let i = topY; i <= botY; i += lineSpace) {
-          Glyph.renderGlyph(ctx, x + this.xShift - 4, i, this.renderOptions.fontScale, 'vexWiggleArpeggioUp');
-        }
+      const lineGlyph =
+        arrow === '\ueb7c' /*arrowheadBlackDown*/ ? '\ueaa9' /*wiggleArpeggiatoUp*/ : '\ueaaa'; /*wiggleArpeggiatoDown*/
```
ronyeh, line 165: What are we trying to do here? Can we document the reasoning with a comment?

src/strokes.ts
```diff
@@ -130,114 +130,87 @@ export class Stroke extends Modifier {
...
+      let txt = '';
+      const el = new Element();
+      // 2 pixels overlap of arpeggiato glyphs
+      for (; el.getWidth() < botY - topY; ) {
+        txt += lineGlyph;
+        el.setText(txt);
+        el.measureText();
+      }
```
ronyeh, line 173: What are we trying to do here? Can we add a comment?

Should this instead be a `while` loop?

src/strokes.ts
```diff
@@ -130,114 +130,87 @@ export class Stroke extends Modifier {
...
     // Draw the rasquedo "R"
     if (this.type === Stroke.Type.RASQUEDO_DOWN || this.type === Stroke.Type.RASQUEDO_UP) {
-      const textFont = Tables.lookupMetricFontInfo(`Strokes.text`);
-      ctx.save();
-      ctx.setFont(textFont);
-      ctx.fillText('R', x + textShiftX, textY);
-      ctx.restore();
+      const el = new Element('Stroke.text');
+      el.setText('R');
+      el.measureText();
```
ronyeh, line 208: Later on, I wonder if it would help to have a cache for measureText() results. We can explore this after your v5 porting process is done. I feel like all the calls to setText() and measureText() will have some negative impact on performance.

ronyeh (changes requested):

src/strokes.ts
rvilarl: And what option is the prefered one? Rasgueado?

rvilarl: OK  already in an issue, no action then,

src/canvascontext.ts
rvilarl: Changed to `angleDegrees`. I decided to go for degrees because this is what SVG uses in rotate.

src/strokes.ts
rvilarl: Documented and selection changed: down for down and up for up

src/strokes.ts
rvilarl: Changed to while with comment.

src/strokes.ts
rvilarl: We have to look into performance, no doubt.

src/strokes.ts
ronyeh: Yeah we might as well follow the wikipedia spelling :-)

src/svgcontext.ts
```diff
@@ -230,6 +230,14 @@ export class SVGContext extends RenderContext {
     this.parent = this.groups[this.groups.length - 1];
   }
 
+  openRotation(angleDegrees: number, x: number, y: number) {
+    this.openGroup().setAttribute('transform', `translate(${x},${y}) rotate(${angleDegrees}) translate(-${x},-${y})`);
```
ronyeh, line 234: I think in CSS, you are supposed to specify the units like 90deg, but does it default to degrees if you omit the unit?

ronyeh (approved):

PR #93, opened 2023-09-08
cleanup - refactor currentMusicFont

rvilarl:
Starting with clean up :) 
Visual changes in formatter  text due to the change in width claculation.

current
Formatter Multiple_Staves___Justified Bravura svg_current
reference
Formatter Multiple_Staves___Justified Bravura svg_reference

### DIFF
--- a/src/stringnumber.ts
+++ b/src/stringnumber.ts
@@ -14,25 +14,9 @@ import { Tables } from './tables';
 import { Category, isStaveNote, isStemmableNote } from './typeguard';
 import { RuntimeError } from './util';
 
-export interface StringNumberMetrics {
-  verticalPadding: number;
-  stemPadding: number;
-  leftPadding: number;
-  rightPadding: number;
-}
-
 export class StringNumber extends Modifier {
   static get CATEGORY(): string {
     return Category.StringNumber;
-  }
-
-  static get metrics(): StringNumberMetrics {
-    return Tables.lookupMetric('StringNumber', {
-      verticalPadding: 0,
-      stemPadding: 0,
-      leftPadding: 0,
-      rightPadding: 0,
-    });
   }
 
   // ## Static Methods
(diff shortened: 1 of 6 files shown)


src/stringnumber.ts
```diff
@@ -27,14 +27,12 @@ export class StringNumber extends Modifier {
...
-        leftPadding: 0,
-        rightPadding: 0,
-      }
-    );
+    return Tables.lookupMetric('StringNumber', {
+      verticalPadding: 0,
+      stemPadding: 0,
+      leftPadding: 0,
+      rightPadding: 0,
+    });
```
ronyeh, line 35: Do we need to pass in a default value here? tables.ts already has a set of values for StringNumber.

tests/formatter_tests.ts
```diff
@@ -62,16 +62,10 @@ const FormatterTests = {
...
-  //
-  // However, tables.ts specifies:
-  //   NOTATION_FONT_SCALE: 39,
-  //   TABLATURE_FONT_SCALE: 39,
-  const musicFont = Tables.currentMusicFont();
-  const glyph: FontGlyph = musicFont.getGlyphs()[glyphName];
-  const widthInEm = (glyph.xMax - glyph.xMin) / musicFont.getResolution();
-  return widthInEm * 38 * Font.scaleToPxFrom.pt;
+  const el = new Element();
+  el.setText(String.fromCharCode(parseInt(glyphName, 16)));
```
ronyeh, line 66: We should default to using String.fromCodePoint() instead, since we are dealing with large unicode code points in our library. Some of the SMuFL codes (e.g., alternates) exceed the limit of fromCharCode().

... Unless we find a performance critical loop where fromCharCode() would help.

tests/formatter_tests.ts
```diff
@@ -556,7 +550,7 @@ function multiStaves(options: TestOptions): void {
   ];
 
   const staveYs = [20, 130, 250];
-  let staveWidth = width + getGlyphWidth('gClef') + getGlyphWidth('timeSig8') + Stave.defaultPadding;
+  let staveWidth = width + getGlyphWidth('E050' /*gClef*/) + getGlyphWidth('E088' /*timeSig8*/) + Stave.defaultPadding;
```
ronyeh, line 553: Since we aren't using the glyph name anymore.... we probably don't need to do the parseInt() and String.fromCharCode() above. Why not just pass in the actual unicode string directly to el.setText()?

rvilarl: reworked

tests/formatter_tests.ts
rvilarl: reworked

src/stringnumber.ts
rvilarl: reworked

ronyeh (approved):

ronyeh:
Looks good, but now there are some merge conflicts to resolve.

rvilarl:
@ron conflict resolved, ready to merge.

PR #94, opened 2023-09-08
Add an enum of SMuFL codes.

ronyeh:
Update one test to use the SMuFL enum.
Note: the enum name is all caps (`SMUFL`) even though the acronym has a lowercase u in the middle.

This PR includes all the SMuFL codes, and adds 92KB to the non-gzipped vexflow.js.

It's not too bad, in my opinion. If anyone ever complains about the size, we can ship a lighter version with only the essential SMuFL codes.


0xfe:
Thanks Ron. This looks good to me. For consistency, maybe we should add overrides (or aliases) for the following:

```javascript
 '4stringTabClef' = '\ue06e', // alias to FourStringTabClef

 '6stringTabClef' = '\ue06d', // alias to SixStringTabClef
```

tests/accidental_tests.ts
```diff
@@ -727,46 +728,52 @@ function sagittal(options: TestOptions): void {
...
+  } = SMUFL;
+
   const notes = [
     f
       .StaveNote({ keys: ['d/4', 'f/4', 'b/4', 'b/4'], duration: '4' })
-      .addModifier(accid('\uE30A' /*accSagittal11MediumDiesisUp*/), 1)
-      .addModifier(accid('\uE303' /*accSagittal5CommaDown*/), 2)
+      .addModifier(accid(accSagittal11MediumDiesisUp), 1)
+      .addModifier(accid(accSagittal5CommaDown), 2)
       .addModifier(accid('b'), 3)
```
rvilarl, line 746: Should `SMUFL` or perhaps `CODES`, as Mohit suggested, also include the vexflow aliases like 'b', '#',...?

ronyeh:
Thanks for the feedback! I'll definitely add the aliases and look into adding the vexflow short codes. Since it won't be exclusively SMuFL codes anymore, we can rename the enum to `Codes`. Or should it be `GlyphCodes`?

I'll let you all know when it's ready for a second review.

tests/accidental_tests.ts
ronyeh: Hmmm... I investigated this and I'm not going to add the vexflow shorthand codes in this PR. 

The short codes in different classes map to different unicode characters.

tables.ts:
```
  static unicode = {
    // ♯ accidental sharp
    sharp: '\u266f',
    // ♭ accidental flat
    flat: '\u266d',
    // ♮ accidental natural
    natural: '\u266e',
    // △ major seventh
    triangle: '\u25b3',
    // ø half-diminished
    'o-with-slash': '\u00f8',
    // ° diminished
    degrees: '\u00b0',
    // ○ diminished
    circle: '\u25cb',
  };
```

tables.ts:
```
const accidentals: Record<string, string> = {
  '#': '\ue262' /*accidentalSharp*/,
  '##': '\ue263' /*accidentalDoubleSharp*/,
  b: '\ue260' /*accidentalFlat*/,
  bb: '\ue264' /*accidentalDoubleFlat*/,
  n: '\ue261' /*accidentalNatural*/,
  '{': '\ue26a' /*accidentalParensLeft*/,
  '}': '\ue26b' /*accidentalParensRight*/,
  db: '\ue281' /*accidentalThreeQuarterTonesFlatZimmermann*/,
  d: '\ue280' /*accidentalQuarterToneFlatStein*/,
  '++': '\ue283' /*accidentalThreeQuarterTonesSharpStein*/,
  '+': '\ue282' /*accidentalQuarterToneSharpStein*/,
  '+-': '\ue446' /*accidentalKucukMucennebSharp*/,
  bs: '\ue442' /*accidentalBakiyeFlat*/,
  bss: '\ue440' /*accidentalBuyukMucennebFlat*/,
  o: '\ue461' /*accidentalSori*/,
  k: '\ue460' /*accidentalKoron*/,
  bbs: '\ue447' /*accidentalBuyukMucennebSharp*/,
  '++-': '\ue447' /*accidentalBuyukMucennebSharp*/,
  ashs: '\ue447' /*accidentalBuyukMucennebSharp*/,
  afhf: '\ue447' /*accidentalBuyukMucennebSharp*/,
};
```

chordsymbol.ts:
```
  static readonly glyphs: Record<string, string> = {
    diminished: '\ue870' /*csymDiminished*/,
    dim: '\ue870' /*csymDiminished*/,
    halfDiminished: '\ue871' /*csymHalfDiminished*/,
    '+': '\ue872' /*csymAugmented*/,
    augmented: '\ue872' /*csymAugmented*/,
    majorSeventh: '\ue873' /*csymMajorSeventh*/,
    minor: '\ue874' /*csymMinor*/,
    '-': '\ue874' /*csymMinor*/,
    '(': '\u0028' /*csymParensLeftTall*/,
    leftParen: '\u0028' /*csymParensLeftTall*/,
    ')': '\u0029' /*csymParensRightTall*/,
    rightParen: '\u0029' /*csymParensRightTall*/,
    leftBracket: '\ue877' /*csymBracketLeftTall*/,
    rightBracket: '\ue878' /*csymBracketRightTall*/,
    leftParenTall: '\u0028' /*csymParensLeftVeryTall*/,
    rightParenTall: '\u0029' /*csymParensRightVeryTall*/,
    '/': '\ue87c' /*csymDiagonalArrangementSlash*/,
    over: '\ue87c' /*csymDiagonalArrangementSlash*/,
    '#': '\ued62' /*csymAccidentalSharp*/,
    b: '\ued60' /*csymAccidentalFlat*/,
  };
```

So the letter `b` could map to `csymAccidentalFlat` or `accidentalFlat`. 

Which one would we put in this enum?

Since each class has its own shorthand, we should probably leave them in each class.

ronyeh (commented):
OK, I think this is ready for review.

I converted `clef.ts` and `timesignature.ts` so we can see what the usage of this enum looks like. (We can continue converting the rest of the unicode uses in future PRs.)

Let me know what you think. I renamed it Glyphs instead of SMUFL, since we might add some other glyphs into this enum (via VexFlow short hand codes).

rvilarl (approved):

tests/accidental_tests.ts
AaronDavidNewman: In vexflow 4.x 'csymParensLeftTall' was the name of the glyph.  So if you wanted to get glyph metrics you would need that, but the 'friendly' name was 'leftParen'.  Since the glyph no longer exists, the friendly name makes more sense.

csymAccidentalFlat was never in the 4.x, that I can remember.  Not sure where that came from.

PR #97, opened 2023-09-10
accidental removed from keyprops

rvilarl:
No visual differences


ronyeh:
What was the purpose of this field? Was it only used during the v4=>v5 conversion?

rvilarl:
I think that this property was also not used in v4

Issue #86, opened 2023-09-02
First render is blotchy (font loading issue?)


AaronDavidNewman:
This might have something to do with lazy-loading the fonts.  I am using font face, as below, and starting the application after on `DOMContentLoaded` event.  It does this consistently on the first load after refreshing the browser, whether or not cache is disabled.  Subsequent loads render fine.

![image](https://github.com/vexflow/vexflow/assets/5438280/c946112d-619e-4fec-bd18-769a8fc1c72a)

``` css
@font-face {
  font-family: 'Bravura';
  src:
    url('fonts/Bravura_1.392.woff') format('woff');
  font-weight: normal;
  font-style: normal;
  font-display: block;
}
```

rvilarl:
@ronyeh you have a solution for that, I believe.

ronyeh:
I haven't had a chance to look into this. I don't think I have experienced this, so is it platform dependent?

Aaron are you on Windows? I'm on Mac and I believe Rodrigo is on some flavor of Linux.

If you have a 100% repro on a test page, can you share the url so I can see if I can see the same thing?

AaronDavidNewman:
It happens if you don't load the fonts programmatically and asynchronously.   i.e. if you just rely on font-face css.  If we don't care about that we can close this.  I was able to work around it by awaiting on the first font load before starting the rendering loop.

ronyeh:
@font-face CSS rules are definitely important. Is there a small test page you can put up that reproduces this problem?

Is it because of the font swapping? Maybe you need to ensure that the font is preloaded (via `<link>` font preloading) before calling any VexFlow methods. I'm curious what causes the issue.

rvilarl:
@ronyeh I have this problem in vexflow-examples

rvilarl:
@AaronDavidNewman  thanks, you are right fonts have to be awaited

PR #98, opened 2023-09-11
Improve Font API

ronyeh:
`vexflow-core.js` is the lightest build with zero fonts. This is the recommended build to use if you want the most flexibility re: loading of SMuFL fonts on the fly.

For continuity with V4:

`vexflow.js` bundles Bravura, Gonville, and Petaluma (and their associated text fonts).
`vexflow-debug.js` is based on vexflow.js, but is not minified.
`vexflow-debug-with-tests.js` is based on vexflow-debug.js and includes all the test cases.
`vexflow-bravura.js` bundles Bravura and Academico. This demonstrates how to bundle a single music font, and can serve as a template if the developer wants to create a `vexflow-gonville.js` (e.g., classic VexFlow).


ronyeh:
Other than the breaking changes, the most common issue I've run into during cleanup is the empty boxes. :-)

image

We should figure out a way to have the first-time developer experience be "free of empty boxes."

It doesn't matter if this means the initial VexFlow download is slightly bigger/slower. For example, we could have VexFlow automatically download the Bravura WOFF2 file out of the box. However, if the developer knows what they are doing re: font preloading, they can set a flag to skip this auto-download. This will avoid the problem of having Bravura downloaded twice on a single web app.

ronyeh:
One approach is to use: `document.fonts.check("16pt Bravura")` to see if Bravura is available to use. If not, we automatically trigger a download of Bravura from jsdelivr.

We can add a configuration option to skip this auto-download. Maybe something like:

```ts
Vex.Flow.skipAutomaticFontCheckAndDownload();
```

Haha... maybe we can figure out a shorter method name or configuration flag.

0xfe:
I strongly support having a zero-cost first-time developer experience. Our default distribution should bundle a font (without any runtime downloads.) We can always have additional minimal distributions with no fonts included.

Optimizing for size here is definitely the wrong tradeoff.

rvilarl:
My idea was to:
- have the download directly from the node_modules directory vexflow_fonts is already a development dependency.
- have the download code outside the vexflow library dierct in html in flow.html and the headless variant.

ronyeh:
> * have the download directly from the node_modules directory. vexflow_fonts is already a development dependency.
> * have the download code outside the vexflow library. Direct in html in flow.html and the headless variant.

The current version of flow.html already loads the font files from the node_modules directory, so there should be no problem when running the flow.html tests.

However, Mohit and I are concerned with making the first time developer experience as fool proof as possible. The library needs to (almost) never have a situation where the developer (or end user) sees those empty square glyphs.

The assumptions I am making:
* The first-time developer does not have any SMuFL fonts installed on his/her operating system.
* The end user who visits that developer's web app also does not have any SMuFL fonts installed.

So when a first-time developer clones this repo, and runs `grunt`, it builds a `vexflow.js`. This library works fine with flow.html, because flow.html handles the font preloading (from node_modules). 

**However**, this library does NOT work on any other HTML test page, or when uploaded to a static server, or when uploaded to a CDN and referenced via JSFiddle.

By default, the library requires the developer to take an extra step, which is to add logic to preload their preferred fonts (e.g., Bravura & Academico).

This will most likely frustrate the developer, even if we have lots and lots of documentation saying that you need to preload a font.

So we are proposing that the default build of vexflow.js either:
1) automatically downloads Bravura & Academico from the jsdelivr CDN, or
2) bundles Bravura & Academico into the JS file (as a base64 encoded).

That way, the default build will show glyphs even if the developer and their users don't have any SMuFL fonts installed.


Later on, if the developer becomes comfortable with VexFlow and reads our README in more detail, they'll discover that grunt also builds a `vexflow-core.js` which bundles zero fonts. Then they can handle the preloading themselves, and call setMusicFont() with the correct fontFamily names.

rvilarl:
@ronyeh Have you look into #86?

0xfe:
I'd vote for (2) for the default build -- principle of least surprise. :-)

rvilarl:
It is strange to spend so much effort to get the glyphs out of the build to then put it back in. But ok let us see what comes out.

ronyeh:
Well, the biggest benefit of your hard work on V5 is that we can now use any SMuFL font with minimal extra work. The vexflow.js version that contains zero fonts will be able to load any *.woff2 or *.otf file on the fly. It's super awesome.

Mohit and I are both concerned that the first time developer experience will be confusing. It's already hard enough to git clone a repo and build it. If the developer throws together a quick test page with the JS file and sees empty squares, they might just give up and look elsewhere. We're trying to make it as easy as possible to adopt VexFlow 5. 

Of course, there will be developers who want the slimmest fastest build possible. You, Simon, and even Michael have noted that it doesn't make sense to bundle Bravura (~300KB) in the library if their web app has already preloaded it for other purposes (e.g., through @font-face CSS rules). 

We willl definitely have a version that supports this goal. Zero bundled fonts. Works with every SMuFL font.

So here's a proposal. We make two builds by default:
1) **vexflow-bravura.js**
2) **vexflow.js**

**Option 1** bundles Bravura, Academico, and Academico Bold. We are not bundling the paths like in V4. Instead, we are bundling the WOFF2 files (as base64) so we can use Rodrigo's new approach. No more need to draw glyph outlines with the Glyph class. Our JSFiddle demos will use this build, because it is a single "do it all" CJS file. You can just stick it in a `<script>` tag and it'll do the right thing.

**Option 2** includes zero fonts. Our README will explain the various options a developer can take to preload their desired SMuFL font for an optimal end user experience.

There is no option 3. We will not be making `vexflow-gonville.js` or `vexflow-petaluma.js` or `vexflow-<font_name>.js` builds by default. However, we might include enough code & comments in the build scripts that will teach a developer how to do so, if they really really want to have a vexflow build that bundles Petaluma.

Does this work for everyone?

0xfe:
That proposal works for me, Ron. How about we call them `vexflow-bravura.js` and `vexflow-minimal.js` (or something.) And hold off on what goes into `vexflow.js`. I'm worried that users of older versions will have more expectations from that last one.

And, of course, thanks for all the improvements, Rodrigo -- we definitely don't want to diminish the value of that. Simply the possibility of having a minimal font-free vexflow for advanced users is a big win.

ronyeh:
> That proposal works for me, Ron. How about we call them `vexflow-bravura.js` and `vexflow-minimal.js` (or something.) And hold off on what goes into `vexflow.js`. I'm worried that users of older versions will have more expectations from that last one.
> 
> And, of course, thanks for all the improvements, Rodrigo -- we definitely don't want to diminish the value of that. Simply the possibility of having a minimal font-free vexflow for advanced users is a big win.

We can stick with the VexFlow 4 convention:
* `vexflow-core.js` is the minimal build with zero bundled fonts.
* `vexflow-bravura.js` only bundles Bravura and Academico (the text font that pairs with Bravura).
* `vexflow.js` ??? In V4, this build bundled Bravura, Gonville, and Petaluma. We can do something similar... but this time I'd recommend only Bravura + Petaluma (and their associated text fonts, Academico and Petaluma Script). 

Honestly... it's not too hard to also build `vexflow-gonvillesmufl.js` and `vexflow-petaluma.js` and others.... but I think we should avoid building versions for every single SMuFL font. We can always add very clear comments in the Gruntfile.js explaining how to do this if you desire.

ronyeh:
After thinking this through, I really think the developer experience needs to be super easy (especially for devs new to VexFlow).

So I propose:
1) `grunt` builds `vexflow.js` and `vexflow-core.js`, along with the debug builds. `vexflow.js` will bundle two fonts (Bravura & Petaluma). This continues somewhat the tradition that VexFlow works out of the box, and can handle both a regular music font (Bravura) and a handwritten music font (Petaluma). `vexflow-core.js` bundles zero fonts, but can load any SMuFL font, thanks to Rodrigo's work.

2) `grunt all` builds both of the above, **plus** individual versions for each SMuFL font that we have tested. So it will build `vexflow-bravura.js`, `vexflow-petaluma.js`, `vexflow-gonvillesmufl.js`, `vexflow-leland.js`, `vexflow-gootville.js`, etc.... If a developer runs `grunt all`, he/she should expect the build to take a while, but they will end up with a custom vexflow build of their choice that bundles only one music font.

The NPM release will NOT include the `grunt all` build products. This requires a developer to clone the repo and run the build themselves.

The NPM release WILL include `vexflow.js` (Bravura & Petaluma), `vexflow-core.js`, `vexflow-debug.js`, `vexflow-debug-with-tests.js`... just like with our V4 release.

We are already breaking enough things in V5. I think we should carry on the tradition of providing a `vexflow.js` that works out of the box, to make the transition slightly easier for any devs upgrading from V4 => V5.

ronyeh:
OK, I was able to fix the jsdom version of the generate_images script.

You need to call node canvas's `registerFont()` with the OTF or WOFF file.

Something like this:
```
const { registerFont } = require('canvas');

// JSDOM uses node-canvas internally. https://www.npmjs.com/package/canvas
// node-canvas can load custom fonts with registerFont(...).
// node-canvas doesn't seem to support WOFF2. Regular WOFF works fine.
const fontsDir = path.resolve(__dirname, '../node_modules/@vexflow-fonts/');
registerFont(path.join(fontsDir, 'bravura/bravura.otf'), { family: 'Bravura' });
registerFont(path.join(fontsDir, 'academico/academico.otf'), { family: 'Academico' });
registerFont(path.join(fontsDir, 'petaluma/petaluma.otf'), { family: 'Petaluma' });
registerFont(path.join(fontsDir, 'petalumascript/petalumascript.otf'), { family: 'Petaluma Script' });
registerFont(path.join(fontsDir, 'gonvillesmufl/gonvillesmufl.otf'), { family: 'Gonville' });
```

ronyeh:
Here's a note on how Dorico handles SVG output:

https://steinberg.help/dorico/v2/en/dorico/topics/print_mode/print_mode_graphics_files_fonts_in_pdf_svg_files_embedding_c.html


PDFs contained subset versions of the SMuFL font files.

SVGs have some symbols converted to paths, but other symbols (like text) are just unicode characters, so they will look like boxes if the user doesn't have the font installed.



> PDF Files
> The music and text fonts, and their sub-sets, that are supplied with Dorico Pro are embedded in PDF files during the export. If you open the PDF files on a different computer, they look the same, even if that computer does not have the fonts installed that are used in the document. If you use different fonts, make sure that these permit embedding.
> 
> SVG Files
> SVG (Scalable Vector Graphics) files do not embed fonts directly. Some font characters, such as note heads, articulations, and accidentals, are converted into outlines, so that they do not depend on the font from which they are taken. Other font characters, such as time signature and tuplet digits, are only encoded using references to the font from which they are taken. The latter also applies to regular text, such as staff labels, tempo instructions, and dynamics. This means that the SVG file looks incorrect if rendered by a web browser on a computer that does not have the fonts installed. The appearance of SVG files depends on the browser or the rendering software, and on the fonts that are installed on the computer.
> 
> To ensure that the SVG file appears correctly if embedded in a web page, you can open the SVG file in an illustration program and convert all font characters to outline paths, then re-export the SVG file and embed that file. Alternatively, you can use web fonts to ensure that the necessary fonts are available on the web server.
> 
> SVG graphics that are exported from Dorico Pro conform to the SVG Tiny 1.1 specification, which defines a subset of features in the full SVG specification.
> 
> For information about using web fonts with SVG, refer to the Help Center on the Steinberg website.

ronyeh:
@rvilarl I see you have been updating the README. Let's wait unti lthis PR is done before we finalize the README, because there will be different options on loading fonts.

`@font-face` is one of them, but I think it might be easier to use the JS API that will be proposed by this PR. I'll let you know when it's done.

I suggest that the README offers the **easiest** way to integrate and test VexFlow. As we discussed with Mohit earlier in this thread, maybe the best way would be to have a vexflow.js build that includes the Bravura font so that we don't have to risk the empty boxes and blurred fonts due to the complications of font loading.

I think font loading should be reserved for developers who would like a more customized experience (with a lighter initial VexFlow download).

Anyways, would you be OK with temporarily rolling back your readme changes and packaging them together into a single PR? We can discuss in your PR thread. That way, the README will match the API proposed in this PR.

ronyeh:
For now, feel free to edit as much as you want on the wiki pages: https://github.com/vexflow/vexflow/wiki

I think other than making releases, neither of us should be committing directly to this repository (even if it's for the README).

Committing directly to the repository makes it difficult to tell what's getting added or removed, because every time I do a fetch or pull, my repo just changes without us having a chance to have looked over the PR diff.

rvilarl:
@ronyeh sorry, I thought that that was the right way to change the README file. I will rollback and make a PR. 

I solved the problem in vexflow-examples. Could you visit https://vexflow.github.io/vexflow-examples/ on your devices and tell me if it works with them?

ronyeh:
Thanks. Yes, I see the examples and they look fine. You should also delete the fonts from your OS to experience the site as a new visitor would. :-)

ronyeh:
This PR continues the V4 tradition of separate build targets, to please everyone. :-)

`vexflow-core.js` is the lightest build (333KB) and contains no fonts.
`vexflow-bravura.js` is 715K and bundles Bravura and Academico (the text font that goes along with Bravura).
`vexflow.js` is 1.1MB but it bundles Bravura, Petaluma, (and will also include Gonville), along with their text fonts.
`vexflow-debug.js` is just an unminified `vexflow.js`
`vexflow-debug-with-tests.js` is `vexflow-debug.js` plus all the test cases that are used in `flow.html`.

All of these builds have the same ability to load and change fonts on the fly via loadFonts() and setFonts().

We provide builds with bundled fonts to avoid the potential pitfalls of seeing empty squares during testing. `vexflow-core.js` is for more advanced users who want to make sure their favorite fonts are lazy loaded, or are loaded no more than once.

Note: in this PR, temporarily ignore anything related to Gonville or GonvilleSmufl. This will be fixed in a future PR once we have an official SMuFL-compliant build of Gonville.

rvilarl:
I am not sure if vexflow.js with now Bravura and Petaluma and later with Gonville is the best approach. The library will be very big. We have also removed the combination of fonts in tests. Perhaps vexflow should have no font and then have individuals vexflow-petaluma, vexflow-gonville,...
 
In any case we can adapt that in the future. I merge now.

ronyeh:
Hi Rodrigo / @rvilarl ,

I took this approach for several reasons.

1) It is the same approach we took in VexFlow 4, where `vexflow.js` bundled some default fonts, and `vexflow-core.js` was a lightweight version that required dynamic loading of fonts.

2) Because it is familiar, developers who have used V4 will not be surprised when upgrading to V5. See these discussion comments earlier in this thread:

> Comment A - "principle of least surprise" https://github.com/vexflow/vexflow/pull/98#issuecomment-1718257686
> Comment B - concern for older users who expect vexflow.js to work out of the box https://github.com/vexflow/vexflow/pull/98#issuecomment-1719342245
> Comment C - my comment supporting the continuation of V4's approach of vexflow.js vs vexflow-core.js - https://github.com/vexflow/vexflow/pull/98#issuecomment-1720285583

3) Using `vexflow.js` (with bundled fonts) as the demo & tutorial build will avoid the "empty squares" problem we have been running into. As a VexFlow maintainer, you have installed all the music fonts onto your operating system, which means that you will likely never run into this issue. However, most developers new to VexFlow will NOT have SMuFL fonts installed on their system. I really want to avoid the case where a developer copies a tutorial HTML file and loads it in their browser and sees a bunch of □□□□□□□□□□□□□□□ on their score.

4) We should not be overly concerned about the size of `vexflow.js`. Our documentation can highlight the benefits of using the lighter build. If a developer has concerns about the size of the library, he/she can use `vexflow-core.js` instead. Here are the sizes as reported by my macOS Finder:

> **`vexflow-core.js` approach** 634KB total
> * vexflow-core.js (341KB)
> * bravura.woff2 (247KB)
> * academico.woff2 (23KB) + academico-bold.woff2 (23KB)
> 
> **`vexflow.js` approach** 1.13MB total
> * vexflow.js (currently 1.1MB, but will grow to ~1.13MB once we add in Gonville)

So yes, it is 500KB more to include Petaluma, Petaluma Script, and Gonville. But that means any **Getting Started** tutorial will work out of the box with three fonts. We could drop Gonville, but it's a tiny font at about 30KB and I figured it'd be fun to keep it in there for old times' sake.

If you want to discuss further, you can open a new GitHub issue and tag all of us and we can iron this out.

PR #99, opened 2023-09-12
textbracket (v5)

rvilarl:
TextBracket ported. Visual differences:

current
TextBracket Simple_TextBracket Bravura svg_current
reference
TextBracket Simple_TextBracket Bravura svg_reference
current
TextBracket TextBracket_Styles Gonville svg_current
reference
TextBracket TextBracket_Styles Bravura svg_reference

### DIFF
--- a/src/textbracket.ts
+++ b/src/textbracket.ts
@@ -50,8 +50,8 @@ export class TextBracket extends Element {
     bracketHeight: number;
   };
 
-  protected readonly text: string;
-  protected readonly superscript: string;
+  protected textElement: Element;
+  protected superscriptElement: Element;
 
   protected line: number;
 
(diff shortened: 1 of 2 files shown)


src/textbracket.ts
```diff
@@ -70,30 +70,19 @@ export class TextBracket extends Element {
...
     this.stop = stop;
 
-    this.text = text;
-    this.superscript = superscript;
+    this.textElement = new Element('TextBracket');
+    this.textElement.setText(text);
+    this.textElement.measureText();
+    this.superscriptElement = new Element('TextBracket');
+    this.superscriptElement.setText(superscript);
+    const smallerFontSize = Font.scaleSize(this.textFont.size, 0.714286);
```
ronyeh, line 84: Just curious where the 0.714... comes from? :-)

src/textbracket.ts
```diff
@@ -119,7 +108,13 @@ export class TextBracket extends Element {
    * @returns this
    */
   applyStyle(ctx: RenderContext): this {
-    ctx.setFont(this.font);
+    this.textElement.setFont(this.textFont);
+    // We called this.resetFont() in the constructor, so we know this.textFont is available.
+    const { family, size, weight, style } = this.textFont;
+    // To draw the superscript, we scale the font size by 1/1.4.
```
ronyeh, line 114: I see, it is 1/1.4. Should this be customizable?

ronyeh:
Looks like we should change the default text font for Bravura to be Academico. :-) I don't think Roboto Slab fits as well alongside this music font.

src/textbracket.ts
rvilarl: It was already present in the original source. I agree, that it is not nice but I wanted to minimize the differences.

src/textbracket.ts
rvilarl: I suggest to make a general review on constants but not as part of this PR. The super and sub indexes should have, I believe, a common ratio (textbrackets, chordsymbols,...)

ronyeh (approved):

ronyeh:
Thanks!

PR #100, opened 2023-09-12
common metrics fonts

rvilarl:
Lots of visual differences due to font changes.

### DIFF
--- a/src/font.ts
+++ b/src/font.ts
@@ -1,3 +1,5 @@
+import { Tables } from "./tables";
+
 export interface FontInfo {
   /** CSS font-family, e.g., 'Arial', 'Helvetica Neue, Arial, sans-serif', 'Times, serif' */
   family?: string;
@@ -117,8 +119,8 @@ export class Font {
       family = f;
     }
 
-    family = family ?? '';
-    size = size ?? '';
+    family = family ?? Tables.lookupMetric('fontFamily') as string;
+    size = size ?? Tables.lookupMetric('fontSize')+'pt';
     weight = weight ?? FontWeight.NORMAL;
     style = style ?? FontStyle.NORMAL;
 
@@ -187,7 +189,7 @@ export class Font {
     let size: string;
     const sz = fontInfo.size;
     if (sz === undefined) {
-      size = '';
+      size = Tables.lookupMetric('fontSize')+'pt';
     } else if (typeof sz === 'number') {
       size = sz + 'pt ';
     } else {
@@ -195,7 +197,7 @@ export class Font {
       size = sz.trim() + ' ';
     }
 
-    const family: string = fontInfo.family ?? '';
+    const family: string = fontInfo.family ?? Tables.lookupMetric('fontFamily');
 
     return `${style}${weight}${size}${family}`;
   }
--- a/src/font.ts
+++ b/src/font.ts
@@ -1,6 +1,3 @@
-import { TupletMetrics } from './tuplet';
-import { defined } from './util';
-
 export interface FontInfo {
   /** CSS font-family, e.g., 'Arial', 'Helvetica Neue, Arial, sans-serif', 'Times, serif' */
   family?: string;
@@ -35,15 +32,6 @@ export class Font {
   //////////////////////////////////////////////////////////////////////////////////////////////////
   // STATIC MEMBERS
 
-  /** Default sans-serif font family. */
-  static SANS_SERIF: string = 'Arial, sans-serif';
-
-  /** Default serif font family. */
-  static SERIF: string = 'Times New Roman, serif';
-
-  /** Default font size in `pt`. */
-  static SIZE: number = 10;
-
   // CSS Font Sizes: 36pt == 48px == 3em == 300% == 0.5in
   /** Given a length (for units: pt, px, em, %, in, mm, cm) what is the scale factor to convert it to px? */
   static scaleToPxFrom: Record<string, number> = {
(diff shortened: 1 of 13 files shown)


src/font.ts
```diff
@@ -129,8 +117,8 @@ export class Font {
       family = f;
     }
 
-    family = family ?? Font.SANS_SERIF;
-    size = size ?? Font.SIZE + 'pt';
+    family = family ?? '';
+    size = size ?? '';
```
ronyeh, line 121: I suggest we hardcode something here as a fallback, if you want to remove the constants.

This `validate()` method is supposed to return a valid font object that makes sense in the CSS world.

So we should put 'sans-serif' as the fallback family, and '10pt' or '12pt' as the fallback font size, so that this method doesn't return empty strings for those fields.

src/font.ts
```diff
@@ -199,15 +187,15 @@ export class Font {
     let size: string;
     const sz = fontInfo.size;
     if (sz === undefined) {
-      size = Font.SIZE + 'pt ';
+      size = '';
```
ronyeh, line 190: Same here. We want to have this method always return a valid CSS font declaration. The default size should be something like '10pt' or '12pt'. Hardcoding it is fine if you don't like the constants.

src/font.ts
```diff
@@ -199,15 +187,15 @@ export class Font {
...
     } else if (typeof sz === 'number') {
       size = sz + 'pt ';
     } else {
       // size is already a string.
       size = sz.trim() + ' ';
     }
 
-    const family: string = fontInfo.family ?? Font.SANS_SERIF;
+    const family: string = fontInfo.family ?? '';
```
ronyeh, line 198: Same here. We want to have this method always return a valid CSS font declaration. The default family should be something like 'Arial, sans-serif' instead of just an empty string.

src/svgcontext.ts
```diff
@@ -155,8 +155,8 @@ export class SVGContext extends RenderContext {
     this.lineWidth = 1.0;
 
     const defaultFontAttributes = {
-      'font-family': Font.SANS_SERIF,
-      'font-size': Font.SIZE + 'pt',
+      'font-family': Tables.lookupMetric('fontFamily'),
+      'font-size': '10pt',
```
ronyeh, line 159: It's interesting that we have a default for fontFamily, but not for fontSize.

Maybe there should be a:

```
Tables.lookupMetric('fontSize'),
```

that returns 10pt?

ronyeh: Wait.... there is already such a field. Doesn't the default look like:

```
  fontFamily: 'Bravura',
  fontSize: 30,
  fontWeight: 'normal',
  fontStyle: 'normal',
```

So instead of '10pt' shouldn't we do a `Tables.lookupMetric('fontSize')` here?

src/tables.ts
```diff
@@ -146,14 +141,12 @@ export const CommonMetrics: Record<string, any> = {
   },
 
   TabSlide: {
-    fontFamily: 'Times New Roman, serif',
```
ronyeh, line 149: So our previous VexFlow contributors have decided that some elements should be `Arial, sans-serif` but other elements should be `Times New Roman, serif`..... By removing all these, does that mean all of the elements will use `Roboto Slab`? (Or Academico, when we switch over.)

ronyeh: Should it be easy to customize? Or at least we can document how to customize it by adding such a fontFamily entry inside the CommonMetrics object.

ronyeh (changes requested):

src/svgcontext.ts
rvilarl: We have to fix some elements with no defaults in CommonMetrics and that expect it to be 10. In the end the default should be 30.

src/tables.ts
rvilarl: This is right.

src/svgcontext.ts
ronyeh: That's why I think there needs to be a separation between music notation fonts and text fonts (for showing letters on the score). 

30pt doesn't make as much sense for the text elements (like volta bracket labels). 

Similarly, setting 'Roboto Slab' on the gClef symbol also doesn't make much sense.

ronyeh: One idea for this would be to have these methods in `flow.ts`:

```ts
  static setMusicFont(...fontNames: string[]): void {
    CommonMetrics.musicfontFamily = fontNames.join(',');
    CommonMetrics.fontFamily = CommonMetrics.musicfontFamily + ',' + CommonMetrics.textfontFamily;
  }

  static getMusicFont(): string[] {
    return Tables.lookupMetric('musicfontFamily').split(',');
  }

  static setTextFont(...fontNames: string[]): void {
    CommonMetrics.textFontFamily = fontNames.join(',');
    CommonMetrics.fontFamily = CommonMetrics.musicfontFamily + ',' + CommonMetrics.textfontFamily;
  }

  static getTextFont(): string[] {
    return Tables.lookupMetric('textFontFamily').split(',');
  }
```

The basic idea would be to have two separate font families (and default font sizes).

If an Element is only music symbols, it will do a lookupMetric on `musicFontFamily`. If an Element is only text characterse, it will lookupMetric('textFontFamily'). If an Element has both, it will lookupMetric('fontFamily'), which is the `${musicFontFamily},${textFontFamily}`.

p.s., i haven't checked the above code. There's probably bugs with this approach (e.g., the joining with a comma).

ronyeh:
Wow this is already PR #100. Lots of hard work being done! :-)

src/svgcontext.ts
rvilarl: @ronyeh this separation does not much the SVG approach. In a SVG you can have one default that will be inherited by the elements inside. At these level I think that what makes more sense is to have the fontFamily including all the fonts, SMuFL and non SMuFL. Have a look at the SVGs produced by this branch, the redefinition of fonts in SVGs is much leaner.

src/font.ts
rvilarl: done, use configured one

src/font.ts
rvilarl: done

src/font.ts
rvilarl: done

ronyeh (approved):

PR #101, opened 2023-09-13
measure text reworked

rvilarl:
Reworlk to reduce the number of calls to measureText. Now only called when a recalculation is required based on #metricValid. No visual differences

current
image
reference
image

### DIFF
--- a/src/crescendo.ts
+++ b/src/crescendo.ts
@@ -58,7 +58,6 @@ export class Crescendo extends Note {
   }
 
   protected decrescendo: boolean;
-  protected height: number;
   protected line: number;
   protected options = {
     // Extensions to the length of the crescendo on either side
--- a/src/element.ts
+++ b/src/element.ts
@@ -88,12 +88,12 @@ export class Element {
 
   protected rendered: boolean;
   protected style?: ElementStyle;
-  protected boundingBox?: BoundingBox;
   protected registry?: Registry;
 
-  protected textFont: Required<FontInfo>;
-  protected text = '';
-  protected textMetrics: TextMetrics = {
+  #textFont: Required<FontInfo>;
+  #text = '';
+  #metricsValid = false;
+  #textMetrics: TextMetrics = {
     fontBoundingBoxAscent: 0,
     fontBoundingBoxDescent: 0,
     actualBoundingBoxAscent: 0,
@@ -103,8 +103,8 @@ export class Element {
     width: 0,
   };
 
-  protected height: number = 0;
-  protected width: number = 0;
+  #height: number = 0;
+  #width: number = 0;
   protected xShift: number = 0;
   protected yShift: number = 0;
   protected x: number = 0;
@@ -118,7 +118,7 @@ export class Element {
     };
 
     this.rendered = false;
-    this.textFont = Tables.lookupMetricFontInfo(this.#attrs.type);
+    this.#textFont = Tables.lookupMetricFontInfo(this.#attrs.type);
 
     // If a default registry exist, then register with it right away.
     Registry.getDefaultRegistry()?.register(this);
@@ -308,8 +308,8 @@ export class Element {
   }
 
   /** Get the boundingBox. */
-  getBoundingBox(): BoundingBox | undefined {
-    return this.boundingBox;
+  getBoundingBox(): BoundingBox {
+    return new BoundingBox(0, -this.textMetrics.actualBoundingBoxAscent, this.width, this.height);
   }
 
   /** Return the context, such as an SVGContext or CanvasContext object. */
@@ -341,7 +341,7 @@ export class Element {
 
   /** Returns the CSS compatible font string for the text font. */
   get font(): string {
-    return Font.toCSSString(this.textFont);
+    return Font.toCSSString(this.#textFont);
   }
 
   /**
@@ -366,25 +366,25 @@ export class Element {
     const fontIsString = typeof font === 'string';
     const sizeWeightStyleAreUndefined = size === undefined && weight === undefined && style === undefined;
 
+    this.#metricsValid = false;
     if (fontIsObject) {
       // `font` is case 1) a FontInfo object
-      this.textFont = { ...defaultTextFont, ...font };
+      this.#textFont = { ...defaultTextFont, ...font };
     } else if (fontIsString && sizeWeightStyleAreUndefined) {
       // `font` is case 2) CSS font shorthand.
-      this.textFont = Font.fromCSSString(font);
+      this.#textFont = Font.fromCSSString(font);
     } else {
       // `font` is case 3) a font family string (e.g., 'Times New Roman').
       // The other parameters represent the size, weight, and style.
       // It is okay for `font` to be undefined while one or more of the other arguments is provided.
       // Following CSS conventions, unspecified params are reset to the default.
-      this.textFont = Font.validate(
+      this.#textFont = Font.validate(
         font ?? defaultTextFont.family,
         size ?? defaultTextFont.size,
         weight ?? defaultTextFont.weight,
         style ?? defaultTextFont.style
       );
     }
-    this.measureText();
     return this;
   }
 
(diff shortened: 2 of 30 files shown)


ronyeh:
Not sure I'm reading the screenshots right. Is the current version slower?

ronyeh:
Oh I see... only looking at measureText() the current is faster. But I guess it's random noise that the total time taken is longer? Or did something else cause the increase in total time?

src/font.ts
```diff
@@ -261,44 +261,6 @@ export class Font {
...
-    FinaleAsh: 'finale/FinaleAsh_1.7.woff2',
-    FinaleAshText: 'finale/FinaleAshText_1.3.woff2',
-    FinaleJazz: 'finale/FinaleJazz_1.9.woff2',
-    FinaleJazzText: 'finale/FinaleJazzText_1.3.woff2',
-    FinaleBroadway: 'finale/FinaleBroadway_1.4.woff2',
-    FinaleBroadwayText: 'finale/FinaleBroadwayText_1.1.woff2',
-    FinaleMaestro: 'finale/FinaleMaestro_2.7.woff2',
-    FinaleMaestroText: 'finale/FinaleMaestroText-Regular_1.6.woff2',
-  };
-
```
ronyeh, line 301: I am curious why the ability to load web fonts from CDN was deleted?

I thought you wanted to have a vexflow build that bundles zero fonts. To do that, we should have easy APIs for people to call to load their desired font. I'm in the middle of improving this font loading, but I don't think it should be removed altogether.

src/font.ts
```diff
@@ -316,31 +278,4 @@ export class Font {
...
-    }
-
-    const host = Font.WEB_FONT_HOST;
-    for (const fontName of fontNames) {
-      const fontPath = allFiles[fontName];
-      if (fontPath) {
-        Font.loadWebFont(fontName, host + fontPath);
-      }
-    }
-  }
```
ronyeh, line 345: Same as above. I don't see why we need to remove the ability to loadWebFonts() from a CDN.

I think having an easy API to load a desired font is important. Otherwise, a developer will have to look through our documents to understand how to deploy their own fonts on their own preferred CDN.

ronyeh:
I haven't had a chance to review this, but I have an early question about the removal of font loading APIs.

src/font.ts
rvilarl: That was just a check to see that it was not usted. I will revert it.

ronyeh: ? :-) You could have asked haha. I appreciate that you are going through and removing deprecated & old stuff that no may longer apply, but:

1) this deletion isn't related to this PR.

2) isn't it the job of our library to make it easier for developers to use SMuFL fonts? If we remove all the font loading code from the library, every developer will have to figure it out for themselves. You've spent a lot of time on this code base so the idea of preloading a SMuFL font seems easy, but I think there are lots of developers who would appreciate having an easy API call to load a web font (even if they end up rolling their own custom solution later on).

rvilarl: @ronyeh I left that by mistake after a try. I think though that leaving loadWebFont is enough, the user should be able to decide on cdn and font. Anyway as you said, no part of this. :)

rvilarl: reverted

src/font.ts
rvilarl: reverted

rvilarl:
@ronyeh you are right some functions are slower and some functions are quicker but the profiling was based on measureText which is now quicker.

Main changes:
- attributes related to measureText are now private in Element
- accessors are now available. These take into account when measureText has to be called (it is no longer needed to call measureText on the derived classes or anywhere in the code outside Element, it could be a private function)
- textFont is now fontInfo on the derved classes. This accessor was already available.
- boundingBoxes is no longer a member variable, it is calculated in runtime.

ronyeh:
What's the real life impact on flow.html running time? Does it save a noticeable amount of time running the tests?

rvilarl:
> What's the real life impact on flow.html running time? Does it save a noticeable amount of time running the tests?

No it does not, no overall impact, but resolves the problem of keeping the metrics, width,... consistent when you change the text or the font of an element. It also makes sure that measureText is called only when required.

src/crescendo.ts
```diff
@@ -58,7 +58,6 @@ export class Crescendo extends Note {
   }
 
   protected decrescendo: boolean;
-  protected height: number;
```
ronyeh, line 61: I see height is removed in a few classes. Do we not need this anymore?

src/element.ts
```diff
@@ -88,12 +88,12 @@ export class Element {
 
   protected rendered: boolean;
   protected style?: ElementStyle;
-  protected boundingBox?: BoundingBox;
   protected registry?: Registry;
 
-  protected textFont: Required<FontInfo>;
-  protected text = '';
-  protected textMetrics: TextMetrics = {
+  #textFont: Required<FontInfo>;
```
ronyeh, line 93: I thought you renamed textFont to fontInfo in other classes. Why not this one?

ronyeh (changes requested):
I won't have time to do a full review until Sunday or Monday, I think. Here are some quick questions.

rvilarl:
The difference is not significant they also differ from run to run. This is on an older computer, the new one is broken :( 
reference(main)
image
current
image

tests/mocks.ts
```diff
@@ -15,7 +15,6 @@ class MockTickable extends Tickable {
   ticks: Fraction = new Fraction(1, 1);
   voice?: Voice;
   stave?: Stave;
-  width: number = 0;
```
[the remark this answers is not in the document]
rvilarl: Mock derives from Element, which already defines width.

src/crescendo.ts
rvilarl: These classes derive from Element which already define height (this is like width above). The compiler was not complaining with variables but it does complain when the items are accessors in the base class.

src/element.ts
rvilarl: Well, I left it as it is. Element had the internal protected variable textFont and two accessors defined one fontInfo returning FontInfo and one font returning string. I changed textFont to private and used the already existent accessor fontInfo.

src/system.ts
```diff
@@ -295,10 +295,14 @@ export class System extends Element {
     this.startX = startX;
     this.debugNoteMetricsYs = debugNoteMetricsYs;
     this.lastY = y;
-    this.boundingBox = new BoundingBox(this.options.x, this.options.y, this.options.width, this.lastY - this.options.y);
     Stave.formatBegModifiers(this.partStaves);
   }
 
+  /** Get the boundingBox. */
+  getBoundingBox(): BoundingBox {
+    return new BoundingBox(this.options.x, this.options.y, this.options.width, this.lastY ?? 0 - this.options.y);
```
ronyeh, line 303: 0 - this.options.y

Can't it just be -this.options.y

src/element.ts
```diff
@@ -393,14 +393,14 @@ export class Element {
...
+    return Font.toCSSString(this.#textFont);
   }
 
   /** Return a copy of the current FontInfo object. */
   get fontInfo(): Required<FontInfo> {
     // We can cast to Required<FontInfo> here, because
-    // we just called resetFont() above to ensure this.textFont is set.
-    return this.textFont;
+    // we just called resetFont() above to ensure this.fontInfo is set.
+    return this.#textFont;
```
ronyeh, line 403: If you're going to rename textFont in the subclasses... might as well do it for Element. It's a private var anyways.

`this.#textFont` => `this.#fontInfo`

src/element.ts
ronyeh: Might as well have the private instance var match the accessor. Since you no longer want to differentiate between text fonts vs music fonts.... there is no need for it to be called "textFont".

ronyeh (changes requested):
See most recent comments.

src/system.ts
ronyeh: Oh wait... is there a operator precedence thing going on? Do we always want to subtract this.options.y?

rvilarl: this.lastY is normally not undefined. So it is this.lasty - this.options.y

ronyeh: Anyways, I think `this.lastY ?? 0 - this.options.y` is confusing. Not sure what the code is trying to do....

I think you're trying to say:

`this.lastY`

OR if the above is `undefined`, then 

`-this.options.y`

Correct?

ronyeh: Make sure you test the operator precedence here.

Doesn't ?? just result in `this.lastY` if as you say, it is normally defined?

src/element.ts
rvilarl: done

src/system.ts
ronyeh: I think we should add parentheses for clarity, or extract a local variable so it's doing what you actually want it to do.

rvilarl: I will check right now

rvilarl: Precedence is on ??. I have put parentheses to avoid confusion.

rvilarl:
@ronyeh ready, now ?? with parentheses.

ronyeh:
Thanks!

PR #103, opened 2023-09-16
profile - measure text with element

rvilarl:
Looking further into the measure text details, I found that two classes were still using the method from SVGContext.

image

I have reworked it to use element and the good news is that it is quicker withour the need of cache.

image

### DIFF
--- a/src/svgcontext.ts
+++ b/src/svgcontext.ts
@@ -2,6 +2,7 @@
 // MIT License
 // @author Gregory Ristow (2015)
 
+import { Element } from './element';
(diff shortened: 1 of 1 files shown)


rvilarl:
Visual differences are minor. Two examples, current first: 

TabTie Pulloffs Bravura svg_current
TabTie Pulloffs Bravura svg_reference


Vibrato Vibrato_with_Bend Bravura svg_current
Vibrato Vibrato_with_Bend Bravura svg_reference

ronyeh (approved):

PR #106, opened 2023-09-18
tables access internal

rvilarl:
@ronyeh as you suggested CommonMetrics is now on a different file as Metrics. I have also provided access to UNISON in flow to avoid the usage of Tables in the tests.

No visual differences.


src/accidental.ts
```diff
@@ -61,10 +62,10 @@ export class Accidental extends Modifier {
     // If there are no accidentals, no need to format their positions.
     if (!accidentals || accidentals.length === 0) return;
 
-    const noteheadAccidentalPadding = Tables.lookupMetric('Accidental.noteheadAccidentalPadding');
+    const noteheadAccidentalPadding = Metrics.lookupMetric('Accidental.noteheadAccidentalPadding');
```
ronyeh, line 65: Can we rename to the shorter version:

`Metrics.lookup(...)`

ronyeh: Or... how about

```
Metrics.get(...)
Metrics.set(...)
```

It's short and easy to understand. If we don't want to implement set for now, that's fine. But I can see a situation where a developer might want to temporarily change the metrics for a certain group of music symbols.

src/flow.ts
```diff
@@ -241,6 +242,14 @@ export class Flow {
...
   }
 
+  static get UNISON(): boolean {
+    return Tables.UNISON;
+  }
+
+  static set UNISON(unison: boolean) {
+    Tables.UNISON = unison;
+  }
+
```
ronyeh, line 252: Interesting. What does this flag do? It is only referenced on one line in the repo. Maybe in a different PR we can figure out which class should store this flag? Is it Metrics related, or is it StaveNote related?

src/metrics.ts
```diff
@@ -0,0 +1,212 @@
...
+      if (keyPart) {
+        curr = curr[keyPart]; // Go down one level.
+      } else {
+        break;
+      }
+    }
+
+    return retVal;
+  }
+}
```
ronyeh, line 51: Should there be an exra new line in here? Just checking to see if eslint would recommend the empty line.

src/tables.ts
```diff
@@ -555,52 +555,6 @@ export class Tables {
     return clefs[clef];
   }
 
-  /** Use the provided key to look up a FontInfo in CommonMetrics. **/
-  static lookupMetricFontInfo(key: string): Required<FontInfo> {
```
ronyeh, line 559: Do you also plan to delete the big `CommonMetrics` from tables.ts?

src/flow.ts
ronyeh: I guess Flow is fine for now....

ronyeh (changes requested):
Thanks!

src/tables.ts
rvilarl: That was the aim, and I forgot :). Done

src/metrics.ts
rvilarl: done

src/flow.ts
rvilarl: Left unmodified

src/accidental.ts
rvilarl: Now MetricsDefaults is accesible, but a set would be easier. I have no idea how to code that. Could you do it on another PR?

rvilarl: lookup changed to get.

PR #113, opened 2023-09-25
fix musejazz time signature

### LINKED ISSUE #83: Time signatures in Musejazz.
> Fonts look great. Those time signatures in Muzejazz and Petaluma seem off though.

Petaluma was discussed with the authors and it is supposed to be like that. However I agree Musejazz and also Finale Boradway are off. Thanks for the feedback!

_Originally posted by @rvilarl in https://github.com/vexflow/vexflow/issues/82#issuecomment-1703819116_

rvilarl:
fixes #83

### DIFF
--- a/src/timesignature.ts
+++ b/src/timesignature.ts
@@ -132,7 +132,7 @@ export class TimeSignature extends StaveModifier {
 
     // If the height of the digits is more than two staff spaces (20), shift to the next line
     // in order to center the digits on lines 1 and 5 rather than 2 and 4.
-    this.lineShift = height > 22 ? 1 : 0;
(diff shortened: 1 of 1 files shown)


rvilarl:
current
Formatter Proportional_Formatting___No_Justification MuseJazz svg_current
reference
Formatter Proportional_Formatting___No_Justification MuseJazz svg_reference

ronyeh:
Does this edit fix or break any other fonts? I guess the number doesn't really match the comment above anymore. 

How did you determine this new number? Just trial and error? Just curious if it'll break any other (or future) fonts.

rvilarl:
No it does not break others. Trial and error.

PR #114, opened 2023-09-30
Update demos to work with V5.

ronyeh:
Many of the scripts and pages in `vexflow/demos/` were broken during this V5 migration.

This PR fixes most of them.

However, they depend on a few modifications that will come in the next PR, regarding the measureText(), which might not work in certain contexts (e.g., when the canvas API is not available).

This PR is confined to the `vexflow/demos/*` folder, so it does not impact any of the visual tests.


ronyeh:
This PR was extracted from https://github.com/vexflow/vexflow/pull/98

PR #117, opened 2023-10-04
JSDom Support

ronyeh:
JSDom support. Extracted from https://github.com/vexflow/vexflow/pull/98

To test the jsdom image generation:

```
grunt

VEX_GENERATE_OPTIONS=--backends=jsdom  grunt generate:current
```

It will generate file names like: `Accidental.Automatic_Accidentals___Key_C__Two_Octaves.Finale_Ash.jsdom.png`


ronyeh:
For node canvas apps, we now need to do this to register font files before loading VexFlow:

```
const { createCanvas, registerFont } = require('canvas');

// Node canvas has a different API for loading fonts.
registerFont('../../node_modules/@vexflow-fonts/bravura/bravura.otf', { family: 'Bravura' });
registerFont('../../node_modules/@vexflow-fonts/petaluma/petaluma.otf', { family: 'Petaluma' });

const { Renderer, Stave, StaveNote, Formatter } = Vex.Flow;

const canvas = createCanvas(1000, 500);

// VexFlow 5 uses a hidden canvas for measuring font glyphs.
// https://www.w3.org/TR/2012/WD-html5-author-20120329/the-canvas-element.html#the-canvas-element
// In browsers, canvas elements usually default to 300 x 150.
Vex.Flow.Element.setTextMeasurementCanvas(createCanvas(300, 150));
```

I added `Element.setTextMeasurementCanvas()` because non-browser environments don't have the ability to create a `<canvas>` element for text measurement purposes.

ronyeh:
Question for @rvilarl ... even when we are using the `SVGContext`, we still use a `<canvas>` element to measure the glyphs, correct? Is there any way to measure with SVG? It seems weird to have to use a canvas when we're trying to make SVGs.

rvilarl:
This is right, we use canvas to measure font characters. I am not aware of a svg method to do this. Perhaps there is one.

Issue #118, opened 2023-10-05
vexflow 4 PR 1598


rvilarl:
https://github.com/0xfe/vexflow/pull/1598

RyoSusami:
@rvilarl  I'll try.

PR #119, opened 2023-10-06
Hotfix ornament delayed draw

RyoSusami:
Port from https://github.com/0xfe/vexflow/pull/1598
Issue this repository https://github.com/vexflow/vexflow/issues/118

I tried porting it, but the Y coordinate of the turn of the lower voice went down. Can I assume this is an intended change in VexFlow 5?
This change only affects the X coordinate.

**VexFlow 4**
image

**VexFlow5**
image


rvilarl:
I will look into the y position topic. Thanks a lot!

PR #120, opened 2023-10-06
calculate width without stave

rvilarl:
Allows to get the width without having an assigned stave.

@ronyeh Request from @jaredjj3 to your question in https://github.com/stringsync/vexml/issues/113

### DIFF
--- a/src/keysignature.ts
+++ b/src/keysignature.ts
@@ -165,7 +165,7 @@ export class KeySignature extends StaveModifier {
   }
 
   getWidth(): number {
-    if (!this.formatted) this.format();
+    if (!this.formatted) this.calculateWidth();
 
     return this.width;
   }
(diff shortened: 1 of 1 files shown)


src/keysignature.ts
```diff
@@ -229,6 +229,30 @@ export class KeySignature extends StaveModifier {
     this.formatted = true;
   }
 
+  protected calculateWidth(): void {
```
jaredjj3, line 232: Should this also set `this.formatted = true`, so that if `this.getWidth` is called multiple times, it doesn't keep making `this.width` grow?

jaredjj3:
I really appreciate the fast turnaround! This will make `vexml` much cleaner and use resources more efficiently.

src/keysignature.ts
rvilarl: We cannot set formatted to true. We cannot calculate the shifts without the stave. This is somehow a quick fix. The class deserves a refactoring.

jaredjj3: Ah, I see. Thanks for explaining.

jaredjj3: I also see this.width is being set to 0 at the beginning, so this doesn't seem like it'll cause this.width to grow incorrectly like I first thought.

PR #128, opened 2023-10-14
update readme

rvilarl:
Updates the copyright.
Adresses how to use the library with Font-Face.

I think we should add this as soon as possible to allow other people use it. Then we can include the changes of #98


ronyeh:
OK, this looks good for now.

Although I think the getting started example should not require any font-face CSS. I'd like the developer to be able to copy & paste as little code as possible to make it work on their test page.

PR #129, opened 2023-10-14
npm bump, including release-it.

ronyeh:
I bumped the revisions of our dev dependencies, including the `release-it` package.


package.json
```diff
@@ -89,11 +89,11 @@
...
-    "eslint": "^8.46.0",
+    "concurrently": "^8.2.1",
+    "eslint": "^8.51.0",
     "eslint-config-prettier": "^9.0.0",
-    "eslint-plugin-import": "^2.28.0",
-    "eslint-plugin-prettier": "^5.0.0",
+    "eslint-plugin-import": "^2.28.1",
+    "eslint-plugin-prettier": "^5.0.1",
     "eslint-plugin-simple-import-sort": "^10.0.0",
     "fork-ts-checker-webpack-plugin": "^8.0.0",
```
rvilarl, line 98: fork-ts-checker-webpack-plugin should be 9.0.0
grunt-contrib-qunit 8.0.1

package.json
```diff
@@ -105,23 +105,23 @@
     "grunt-webpack": "^6.0.0",
     "jsdom": "^22.1.0",
     "jspdf": "^2.5.1",
-    "npm": "^9.8.1",
+    "npm": "^9.9.0",
```
rvilarl, line 108: npm should be 10.2.0

package.json
```diff
@@ -105,23 +105,23 @@
...
-    "svg2pdf.js": "^2.2.1",
+    "svg2pdf.js": "^2.2.2",
     "terser-webpack-plugin": "^5.3.9",
-    "ts-loader": "^9.4.4",
+    "ts-loader": "^9.5.0",
     "tsc-watch": "^6.0.4",
     "typedoc": "^0.24.8",
-    "typescript": "^5.1.6",
-    "webpack": "^5.88.2"
+    "typescript": "^5.2.2",
```
rvilarl, line 123: typedoc is no longer required. Are you planning to set up vexflow-docs?

rvilarl (changes requested):
package-lock.json update is missing.

ronyeh:
Thanks for the feedback. I went with a conservative update, which did not automatically move to new major revisions due to concerns with breaking changes.

However, I can try bumping everything to the latest and seeing if there are any problems.

I'll also include the package lock.

package.json
ronyeh: What's the URL of the github pages? Does the 4.x release have its own separate github pages for the API docs?

ronyeh:
@rvilarl I updated to latest, and included package-lock.

flow.html seems to work OK. tests pass.

Now `npm outdated` shows no outdated packages.

ronyeh:
@rvilarl I checked and there are no github pages set up for this repository.  Does your working copy have an extra branch that this repo doesn't have (maybe left over from v4)?

In fact, we need to figure out what to do with vexflow-docs since I think we should have versioned docs for each release (at least each minor 5.x release).

rvilarl (approved):

PR #135, opened 2023-10-18
dot v5

rvilarl:
Dot using font glyph.

Visual differences:

current
Dot Basic Bravura svg_current
reference
Dot Basic Bravura svg_reference

### DIFF
--- a/src/dot.ts
+++ b/src/dot.ts
@@ -2,10 +2,11 @@
 //
 // This class implements dot modifiers for notes.
 
+import { Glyphs } from './glyphs';
 import { Modifier } from './modifier';
 import { ModifierContextState } from './modifiercontext';
 import { Note } from './note';
-import { Category, isGraceNote, isStaveNote, isTabNote } from './typeguard';
+import { Category, isStaveNote, isTabNote } from './typeguard';
 import { RuntimeError } from './util';
 
 export class Dot extends Modifier {
@@ -13,7 +14,6 @@ export class Dot extends Modifier {
     return Category.Dot;
   }
 
-  protected radius: number;
   protected dotShiftY: number;
 
   /** Returns the dots associated to a Note. */
@@ -133,17 +133,13 @@ export class Dot extends Modifier {
 
     this.position = Modifier.Position.RIGHT;
 
-    this.radius = 2;
-    this.setWidth(5);
+    this.setText(Glyphs.augmentationDot);
     this.dotShiftY = 0;
   }
 
   setNote(note: Note): this {
     this.note = note;
-    if (isGraceNote(note)) {
-      this.radius *= 0.5;
-      this.setWidth(3);
-    }
+    this.font = note.font;
     return this;
   }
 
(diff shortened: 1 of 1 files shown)


ronyeh:
The more tightly packed dots seem harder to sight read. Though I'm not an expert.... We need to eventually have some comparisons against professionally engraved music to see how we're doing. Is there any way to adjust the spacing between those dots?

In CSS and canvas, there is a `letterSpacing` option:
https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/letterSpacing

src/dot.ts
```diff
@@ -133,17 +133,13 @@ export class Dot extends Modifier {
...
     this.dotShiftY = 0;
   }
 
   setNote(note: Note): this {
     this.note = note;
-    if (isGraceNote(note)) {
-      this.radius *= 0.5;
-      this.setWidth(3);
-    }
+    this.font = note.font;
```
ronyeh, line 142: So how does it look now for grace notes with a dot?

Is the dot full sized? The original code makes the dot half sized for grace notes.

rvilarl: The note, if it is a grace note, should already be scaled. The recommended scale is not 0,5 but 3/5 or 2/3 I do not remember now.

rvilarl: By the way, what would be the meaning of a dot on a grace note? Do you know if this possible? In any case if a dot is to be presented in a note, I think that it makes sense that they use the same font.

ronyeh: Yeah I have no idea what it would mean to add a dot here. But maybe Grace Note in this case just refers to notes that are rendered in a smaller font (e.g., a voice that is rendered smaller than the main voice)? In any case, I pointed it out just because it was deleted here. I figure it was put in for a good reason at some point. ;-)

ronyeh: I found this jsfiddle from the tutorials: https://jsfiddle.net/ronyeh/smyht3q5/

So apparently you can just stick GraceNotes anywhere, to indicate maybe a separate (less important) voice? What if we wanted to add a dot to those notes? Will the dot be magically scaled down?

rvilarl: @mscuthbert @AaronDavidNewman do dots make sense on grace notes?
@ronyeh I think that they will scale down, I wil add a test if the combination make sense

mscuthbert: Though rare, I've definitely seen grace notes with dots.  Particularly in bagpipe music where you might have a dotted-eighth + 16th + eighth grace figure to show that the figure should be played as quickly as possible before the main note.  The W3C music notation group has been talking about graces a lot recently RE: MNX and basically almost anything you've seen in a score can happen in a grace group (we were talking most recently about the thorny problem of clef changes within a grace group, especially if there are two voices in a measure with different numbers of grace notes at the same time where a clef change happens mid grace group).

Morton Feldman--I believe in _The Viola in My Life_--put a fermata on a grace note.  Once you do that, you can do anything.

rvilarl: @mscuthbert could you provide a picture. I tried to find it but I did not succeed

mscuthbert: https://forums.steinberg.net/t/small-problem-with-embellishments/842546

mscuthbert: And they are frequent in cadenzas which take no metrical time and thus have the same positioning and alignment problems as grace note groups.  (and they can themselves have grace notes on them, such as at the end of the most commonly played cadenza to the Mozart clarinet concerto).

rvilarl: Thanks a lot. And could you provide a picture of the fermata case?

PR #137, opened 2023-10-19
Gonville

ronyeh:
This is pending the rename of the actual font file to Gonville (including all metadata). See https://github.com/rvilarl/gonvillesmufl/issues/1

I then need to publish the new Gonville.woff2 to https://github.com/vexflow/vexflow-fonts and to NPM.


ronyeh:
@rvilarl please try this branch.

Run `grunt` to build.

Then run this command:

```
VEX_GENERATE_OPTIONS='--backends=jsdom --fonts=Gonville,GonvilleSmufl' grunt generate:current
```

It temporarily sets an environment variable to tell the `grunt generate:current` script to use the JSDom backend, and to only test Gonville and GonvilleSmufl.

"Gonville" is Simon's version (you may notice that the origin of flags are fixed.)
"GonvilleSmufl" is your font with the remapped glyphs.

ronyeh:
Also, if you open `flow.html` you will see we are temporarily only testing Bravura, Gonville, and GonvilleSmufl.

I noticed that both Gonville fonts have poor alignment of dots (both augmentation dots and staccato dots need to be adjusted horizontally to the right). At least both Gonville fonts are consistent with each other!

image

So how do we deal with things like this? Not all SMuFL fonts will be exactly the same with regard to bounding boxes for each glyph. Will we need font-specific adjustments / tweaks? I am not sure the default values in `MetricsDefaults` will work perfectly for every font.

rvilarl:
> I noticed that both Gonville fonts have poor alignment of dots (both augmentation dots and staccato dots need to be adjusted horizontally to the right). At least both Gonville fonts are consistent with each other!

good catch this is an origin bug in Gonville and GonvilleSmufl bug. The oirigin is in the center of the Glyph and should be to the left.

> So how do we deal with things like this? Not all SMuFL fonts will be exactly the same with regard to bounding boxes for each glyph. Will we need font-specific adjustments / tweaks? I am not sure the default values in `MetricsDefaults` will work perfectly for every font.

This should not be a reason. As stated above, it is an origin bug. I will document the bugs in this thread and once we get a few we can send it to Simon. ok?

rvilarl:
**augmentationDot origin** x position should be to the left and not centered

Gonville
image

Bravura
image

ronyeh:
Thanks for your help. If indeed these are origin issues, we can fix it with Simon's help.

Once we fix all the bugs with the new Gonville, I will update this PR and remove GonvilleSmufl. Then we can review & merge this PR.

ronyeh:
@rvilarl I have removed GonvilleSmufl from the repo.

This version does not yet include the fixed dot. We will update Gonville with fixes and then submit a separate PR later.

tools/fonts/inspector/glyphs.html
```diff
@@ -99,11 +99,10 @@ <h1 id="title">Glyph Inspector</h1>
           <a onclick="chooseFont('FinaleBroadway')">FinaleBroadway</a> &nbsp;
           <a onclick="chooseFont('FinaleJazz')">FinaleJazz</a> &nbsp;
           <a onclick="chooseFont('FinaleMaestro')">FinaleMaestro</a> &nbsp;
-          <a onclick="chooseFont('GonvilleSmufl')">GonvilleSmufl</a> &nbsp;
+          <a onclick="chooseFont('Gonville')">Gonville</a> &nbsp;
           <a onclick="chooseFont('Gootville')">Gootville</a> &nbsp;
           <a onclick="chooseFont('Gootville Text')">Gootville Text</a> &nbsp;<br />
-          <a onclick="chooseFont('Leipzig')">Leipzig</a> &nbsp;
-          <a onclick="chooseFont('Leland')">Leland</a> &nbsp;
+          <a onclick="chooseFont('Leipzig')">Leipzig</a> &nbsp; <a onclick="chooseFont('Leland')">Leland</a> &nbsp;
```
rvilarl, line 105: This probably should be two lines

rvilarl (approved):

PR #139, opened 2023-10-19
Music + Text Fonts

ronyeh:
Music Fonts are now paired with the recommended Text Fonts.

Tons of visual differences. :-)

I also removed the fallback fonts during testing. For example, I think it is better to know if Leland is missing a glyph. In the past we would just use the Bravura glyph as a fallback.


tests/vexflow_test_helpers.ts
```diff
@@ -208,17 +208,18 @@ export class VexFlowTests {
...
-    'Finale Maestro': ['Finale Maestro', 'Bravura', 'Academico'],
+    Bravura: ['Bravura', 'Academico'],
+    'Finale Ash': ['Finale Ash', 'Finale Ash Text'],
+    'Finale Broadway': ['Finale Broadway', 'Finale Broadway Text'],
+    'Finale Maestro': ['Finale Maestro', 'Finale Maestro Text'],
+    Gonville: ['GonvilleSmufl', 'Academico'],
+    Gootville: ['Gootville', 'Edwin'],
+    Leland: ['Leland', 'Edwin'],
+    MuseJazz: ['MuseJazz', 'MuseJazz Text'],
+    Petaluma: ['Petaluma', 'Petaluma Script'],
```
rvilarl, line 222: Should we add Sebastian and Leipzig?

rvilarl, line 222: Should we add Sebastian and Leipzig?

rvilarl (changes requested):

tests/vexflow_test_helpers.ts
ronyeh: Sure, I can add them. How long does flow.html tests take on your computer? It is pretty slow on my laptop. At some point we will need to figure out an easy way to run only a subset of the tests / subset of the fonts.

rvilarl: I agree that we should probably have the opportunity to select the fonts when testing.

rvilarl: OK we leave it out for now.

PR #141, opened 2023-10-20
Base64 encoded fonts

ronyeh:
This is extracted from https://github.com/vexflow/vexflow/pull/98

To make the transition to V5 easier, we will do the following:
- `vexflow-core.js` will be the version that bundles zero fonts.
- `vexflow.js` will be similar to V4. It will bundle Bravura, Petaluma, Gonville. It will also bundle text fonts Academico and Petaluma Script.
- `vexflow-bravura.js` is a template to teach developers how to bundle their own fonts. In this case, it bundles Bravura and Academico.
- `vexflow-debug.js` is just `vexflow-core.js` but not minified.
- `vexflow-debug-with-tests.js` includes all the test cases, and is used by `flow.html`.

This particular PR contributes the Base64 encoding script. The fonts are written out to `vexflow/src/fonts/`.

In the next PR, we will add the entry files and build targets (similar to V4). 

The benefit of this approach is that it mostly avoids the "empty squares" issue that new developers run into when using vexflow-core.js. But once a developer becomes more comfortable with VexFlow, they can integrate vexflow-core.js to have a lighter initial download, and maximum flexibility with the ability to load any SMuFL font at runtime.


ronyeh:
I also deleted the old font generation scripts. They remain available in the V4 repo: https://github.com/0xfe/vexflow

PR #142, opened 2023-10-20
Fix JSFiddle examples (doctype and meta charset declarations).

### LINKED ISSUE #140: meta charset="UTF-8"
I've noticed in some demos, I get weird symbols if I don't explicitly declare the charset to be **UTF-8**.

```html
  <head>
    <meta charset="UTF-8" />
  </head>
```

Without the declaration, I get this:

image

With the declaration, it looks better:

image

This makes sense, since we are now displaying text instead of rendering paths. It is just something to note. Maybe we need to put these little potential stumbling blocks into a developer's FAQ at some point.

ronyeh:
Addresses #140 by making sure we don't have an empty doctype declaration on example pages.

Also add this block:

```html
  <head>
    <!-- IMPORTANT: Set the character set to UTF-8. Otherwise you may get weird symbols on the score. -->
    <meta charset="utf-8" />
  </head>
```

The charset declaration is important, because when the file is loaded from the file system or from a web server that does not default to UTF-8, the browser may show weird characters on the staff.

### DIFF
--- a/demos/jsfiddle/sandbox.html
+++ b/demos/jsfiddle/sandbox.html
@@ -1,39 +1,46 @@
-<!DOCTYPE >
+<!doctype html>
 <!-- 
   This example is available at:
   https://jsfiddle.net/wjm3t5su/ -->
 <html>
-  <style>
-    body {
-      font-family: Arial, 'sans-serif';
-    }
-  </style>
+  <head>
+    <!-- IMPORTANT: Set the character set to UTF-8. Otherwise you may get weird symbols on the score. -->
+    <meta charset="utf-8" />
+    <style>
+      body {
+        font-family: sans-serif;
+      }
+    </style>
+  </head>
   <body>
     <div id="output"></div>
-    <script src="https://cdn.jsdelivr.net/npm/vexflow/build/cjs/vexflow.js"></script>
+    <!-- <script src="https://cdn.jsdelivr.net/npm/vexflow/build/cjs/vexflow.js"></script> -->
+    <script src="../../build/cjs/vexflow.js"></script>
     <script>
       const { Renderer, Stave } = Vex.Flow;
 
-      // Vex.Flow.setFonts('Bravura');
-      // Vex.Flow.setFonts('Gonville');
-      Vex.Flow.setFonts('Petaluma');
+      Vex.Flow.loadFonts('Bravura', 'Petaluma').then(() => {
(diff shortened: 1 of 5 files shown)


demos/jsfiddle/sandbox.html
```diff
@@ -1,39 +1,46 @@
...
   https://jsfiddle.net/wjm3t5su/ -->
 <html>
-  <style>
-    body {
-      font-family: Arial, 'sans-serif';
-    }
-  </style>
+  <head>
+    <!-- IMPORTANT: Set the character set to UTF-8. Otherwise you may get weird symbols on the score. -->
+    <meta charset="utf-8" />
```
ronyeh, line 8: ¡muy importante!

rvilarl:
@ronyeh this is a duplication of vexflow-examples. The examples there are also following the recommendation from @0xfe to promote the Factory API.

see https://vexflow.github.io/vexflow-examples/

I tried to make a new alpha release to update the font loading to your new load function.

I think that we should remove that here.

ronyeh:
Yeah, I know this is currently duplicating some work, but let's keep it in here for the moment, until vexflow-examples site and content are finalized. 

Currently the vexflow-examples does not highlight the importance of the meta charset tag. It also only shows partial examples (in the style of JSFiddle). 

I think there is a need to have full HTML files that can be copied and pasted and edited. Maybe this example can be eventually moved over to vexflow-examples, but I think having the full HTML page is important.

rvilarl:
Then let us make a release and put the complete HTML head in vexfow-examples. Vexflow-examples also show the text and have a link to jsfiddle. 
This also promotes to use Native API rather than Factory API. See https://github.com/0xfe/vexflow/issues/775

rvilarl:
If you want to merge it go ahead, I have no other comments on the files but I do not see the point.

ronyeh:
No problem. I'm happy to delete these JSFiddle examples.

Let me know when you've added the meta charset to the vexflow-examples. We can adopt loadFonts() / setFonts() after the next alpha release.

ronyeh:
Also regarding Factory API.... :-) I personally prefer the low level API as it is easier for me to understand what's going on. But sure, if we want to promote the Factory API, let's make all our tutorials and documentations focus on it.

rvilarl:
Done https://github.com/vexflow/vexflow-examples/pull/3

@0xfe do you still want to promote the factory API?

PR #143, opened 2023-10-21
Delete JSFiddle examples. 

ronyeh:
Also add `<meta charset="UTF-8">` to remaining demos.


rvilarl (approved):

rvilarl:
@ronyeh will you port the other ones which a possible to port to vexflow-examples or should I do it?

would you make a PR in vexflow-examples to sort the examples in directories?

ronyeh:
Unfortunately I don't have time to help out with the vexflow-examples repo at the moment.

I've been keeping these demos in the same repo because it makes it easy to test and fix. In a separate repo, we would need to make a release before checking if it broke anything.

There's a benefit to having some examples / tests in the main repo.

rvilarl:
If you think that it is a bad idea, we can also delete it and put the demos in the code. I was trying to follow the same approach as per docs (we still did not much in vexflow-docs) and use the github pages facility that allows to seee the examples on line.

mscuthbert:
Just catching up on everything now and wondering how do other devs test out their own version of Vexflow if there isn't a sandbox anymore?  The old (pre-js-fiddle) "Vexflow Sandbox" was always my way of checking that what I was developing had reasonable output without doing a release.

ronyeh:
I think Rodrigo's goal with this was to move most of the examples out into a separate repository:

https://github.com/vexflow/vexflow-examples

I agree there should be a JSFiddle that points to the most recent release (or alpha/beta version on NPM) so that people can test out their code snippets easily.

mscuthbert:
> I agree there should be a JSFiddle that points to the most recent release (or alpha/beta version on NPM) so that people can test out their code snippets easily.

I was asking for Vexflow devs themselves, to test out their own changes to the code before making a PR -- the Sandbox wasn't really an example, it was a meta-tool for testing out examples before committing them.

(here's the music21j sandbox, stolen from the old Vexflow one.
http://cuthbertlab.github.io/music21j/testHTML/m21jSandbox.html
)

PR #144, opened 2023-10-22
metronome musicxml features

rvilarl:
Features missing detected by @jaredjj3 in https://github.com/stringsync/vexml/pull/157

Tests modified to show the new features:

reference
svg_Stave Tempo_Test Bravura svg_reference
current
svg_Stave Tempo_Test Bravura svg_current

### DIFF
--- a/src/stavetempo.ts
+++ b/src/stavetempo.ts
@@ -2,6 +2,7 @@
 // @author: Radosaw Eichler 2012
 
 import { Element } from './element';
+import { Glyphs } from './glyphs';
 import { Metrics } from './metrics';
 import { Stave } from './stave';
 import { StaveModifier, StaveModifierPosition } from './stavemodifier';
@@ -9,10 +10,13 @@ import { Tables } from './tables';
 import { Category } from './typeguard';
 
 export interface StaveTempoOptions {
-  bpm?: number;
+  name?: string;
+  parenthesis?: boolean;
   duration?: string;
   dots?: number;
-  name?: string;
+  bpm?: number;
+  duration2?: string;
+  dots2?: number;
 }
 
 export class StaveTempo extends StaveModifier {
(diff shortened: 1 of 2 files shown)


tests/stave_tests.ts
```diff
@@ -690,8 +690,8 @@ function drawTempo(options: TestOptions, contextBuilder: ContextBuilder): void {
   y += 150;
 
   drawTempoStaveBar(120, { duration: 'w', bpm: 80 }, 0);
-  drawTempoStaveBar(100, { duration: 'h', bpm: 90 }, 0);
-  drawTempoStaveBar(100, { duration: 'q', bpm: 96 }, 0);
+  drawTempoStaveBar(100, { duration: 'h', dots2: 1, duration2: 'q' }, 0);
```
ronyeh, line 693: In the tests / examples, we should probably always put dots after duration for consistency. It looks weird here to have duration / dots2.  It should be duration / duration2 / dots2.

src/stavetempo.ts
```diff
@@ -2,17 +2,21 @@
...
 export interface StaveTempoOptions {
-  bpm?: number;
+  name?: string;
+  parenthesis?: boolean;
   duration?: string;
   dots?: number;
-  name?: string;
+  bpm?: number;
+  duration2?: string;
+  dots2?: number;
```
ronyeh, line 19: Can we add a comment somewhere in this file explaining what duration2 and dots2 is for?

You can explain it is for cases where the tempo marker is X = Y where both X and Y are notes (not BPM numbers).

tests/stave_tests.ts
rvilarl: done

src/stavetempo.ts
rvilarl: done

rvilarl:
@jaredjj3 @ronyeh while documenting I realised that bpm should be a string as the bpm can also be specified, for example, as a range. I have modified the test as well to show a range.

svg_Stave Tempo_Test Bravura svg_current

ronyeh:
Have you pushed the changes? I don't see any of the new work.

rvilarl:
> Have you pushed the changes? I don't see any of the new work.

sorry, now pushed

ronyeh:
Sorry to add more work. 

Doesn't bpm make more sense as a number in almost all cases? The API should maybe accept number | string. String is definitely the less common case. 

Then internally, we could convert all incoming numbers to strings at render time. 

Does that make sense?

It just feels weird to make someone type out '120' with quotes when they just want to set the bpm to 120.

rvilarl:
> Sorry to add more work.
> 
> Doesn't bpm make more sense as a number in almost all cases? The API should maybe accept number | string. String is definitely the less common case.
> 
> Then internally, we could convert all incoming numbers to strings at render time.
> 
> Does that make sense?
> 
> It just feels weird to make someone type out '120' with quotes when they just want to set the bpm to 120.

@ronyeh Done, I think that it is better. Thanks!

ronyeh (approved):

jaredjj3:
Nicely done! Let me know when this is released and I'll weave it into vexml.

rvilarl:
@jaredjj3 already in version alpha.10. Please note that the library without embedded fonts is vexflow-core.js

PR #146, opened 2023-10-25
Vex.Flow.* ⇨ VexFlow.*

ronyeh:
Merge `vex.ts` and `flow.ts` so that we can have a `VexFlow.*` namespace for our methods.


ronyeh:
This is ready for testing.

There are probably still small bugs (especially with the ESM build). But I'd like to get more people testing this branch to verify that it works as expected.

`Vex.Flow.*` is now `VexFlow.*`

ronyeh:
I think we can finish / test / review / merge this before we do the Gonville PR. The Gonville PR is more straightforward.

rvilarl (approved):
I have quickly looked. I think we should merge and resolve issues if any.

Issue #147, opened 2023-10-26
grunt generate:reference fails


rvilarl:
@ronyeh I have realised that after the last merge npm run test:reference fails on the genration of images.

ronyeh:
Ah bummer. I can look into this tomorrow.

PR #148, opened 2023-10-27
Fix `grunt test:reference`

ronyeh:
Fix https://github.com/vexflow/vexflow/issues/147
Also adds Sebastian and Leipzig as test fonts.

### DIFF
--- a/tools/generate_images_jsdom.js
+++ b/tools/generate_images_jsdom.js
@@ -19,6 +19,7 @@ registerFont(path.join(fontsDir, 'bravura/bravura.otf'), { family: 'Bravura' });
 registerFont(path.join(fontsDir, 'academico/academico.otf'), { family: 'Academico' });
 registerFont(path.join(fontsDir, 'petaluma/petaluma.otf'), { family: 'Petaluma' });
 registerFont(path.join(fontsDir, 'petalumascript/petalumascript.otf'), { family: 'Petaluma Script' });
+registerFont(path.join(fontsDir, 'sebastian/sebastian.otf'), { family: 'Sebastian' });
 registerFont(path.join(fontsDir, 'gonvillesmufl/gonvillesmufl.otf'), { family: 'Gonville' });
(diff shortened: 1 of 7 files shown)


tools/generate_images_jsdom.js
```diff
@@ -19,6 +19,7 @@ registerFont(path.join(fontsDir, 'bravura/bravura.otf'), { family: 'Bravura' });
 registerFont(path.join(fontsDir, 'academico/academico.otf'), { family: 'Academico' });
 registerFont(path.join(fontsDir, 'petaluma/petaluma.otf'), { family: 'Petaluma' });
 registerFont(path.join(fontsDir, 'petalumascript/petalumascript.otf'), { family: 'Petaluma Script' });
+registerFont(path.join(fontsDir, 'sebastian/sebastian.otf'), { family: 'Sebastian' });
 registerFont(path.join(fontsDir, 'gonvillesmufl/gonvillesmufl.otf'), { family: 'Gonville' });
 registerFont(path.join(fontsDir, 'finaleash/finaleash.otf'), { family: 'Finale Ash' });
```
rvilarl, line 24: The list is not complete, Leipzig is included in tests but not here (probably to reduce the required time of processing), but sebastian associated text font (Nepomuk) is also not included.

rvilarl:
I will merge to fix the testing. If the comment above is relevant we can make another PR.

tools/generate_images_jsdom.js
ronyeh: This script is for jsdom image generator which we do not use actively. We mostly use the puppeteer version. I didn't have time to test all the edge cases since this was a quick fix. 

We probably need to go through the test helpers and puppeteer + Jason scripts to organize it and maybe even optimize it, since we are testing a whole lot of fonts now.

rvilarl: Agree it  should be possible to select only one and reduce time. Most of the times I only look at Bravura differences.

PR #149, opened 2023-10-27
getElementWidth added

### LINKED ISSUE #76: APIs missed after v5 migration.
@AaronDavidNewman @mscuthbert @sschmidtu please add to this issue broken APIs that you would like to be recoverd.

rvilarl:
fix #76
static function to measure text/glyphs as requested in #76  

No visual differences.

### DIFF
--- a/src/element.ts
+++ b/src/element.ts
@@ -8,7 +8,7 @@ import { Metrics } from './metrics';
 import { Registry } from './registry';
 import { RenderContext } from './rendercontext';
 import { Category } from './typeguard';
-import { defined, globalObject, prefix, RuntimeError } from './util';
+import { defined, prefix, RuntimeError } from './util';
 
 /** Element attributes. */
 export interface ElementAttributes {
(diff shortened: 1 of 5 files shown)


src/element.ts
```diff
@@ -616,6 +616,29 @@ export class Element {
...
+    let txtCanvas: HTMLCanvasElement | OffscreenCanvas | undefined = Element.#txtCanvas;
+    // Create the canvas element that will be used to measure text.
+    if (!txtCanvas) {
+      if (typeof document !== 'undefined') {
+        txtCanvas = document.createElement('canvas'); // Defaults to 300 x 150. See: https://www.w3.org/TR/2012/WD-html5-author-20120329/the-canvas-element.html#the-canvas-element
+      } else if (typeof OffscreenCanvas !== 'undefined') {
+        txtCanvas = new OffscreenCanvas(300, 150);
+      }
+      Element.#txtCanvas = txtCanvas;
+    }
```
ronyeh, line 630: Is this duplicated from Element.measureText()?

If so, we should extract this into a separate method that initializes the Element.#txtCanvas if it does not exist yet.

Maybe `Element.getTextMeasurementCanvas()` and put it right after Element.setTextMeasurementCanvas()? If it is undefined, it creates a new one. If it is not undefined, it just returns the current one.

tests/key_clef_tests.ts
```diff
@@ -22,18 +23,12 @@ const ClefKeySignatureTests = {
...
-function getWidth(code: string) {
-  const el = new Element();
-  el.setText(code);
-  return el.getWidth();
-}
-
 const fontWidths = () => {
-  const sharpWidth = getWidth('\ue262' /*accidentalSharp*/) + 1;
-  const flatWidth = getWidth('\ue260' /*accidentalFlat*/) + 1;
+  const sharpWidth = Element.getElementWidth(Glyphs.accidentalSharp) + 1;
```
ronyeh, line 27: There is already a `Element.getWidth()`. It is unclear how this is different (if I were a developer new to VexFlow).

Should this be called `Element.measureWidth()`?

src/element.ts
rvilarl: done

tests/key_clef_tests.ts
rvilarl: done

PR #156, opened 2023-11-09
Redesign pointer events

### LINKED ISSUE #155: SVG: For text elements, hit-testing is performed on a character cell basis
`return VF.Element.measureWidth(smoGlyph.vexGlyph);`

returns about 86 px for a 'gClef', but the actual width is about 23 px.  All the glyphs have this issue, so the layout ends using too much space.

Was: 
image

Is:
image

Also, the height of the glyphs' bounding box is somehow wrong.  The bounding box returns height of 160 even though the notehead that is being measures is about 14 px.

image

image

image

rvilarl:
The solution is to set a transparent pointer rect to capture the events.

No visual differences but right behaviour on the interactive mouseover tests.

### DIFF
--- a/src/svgcontext.ts
+++ b/src/svgcontext.ts
@@ -5,7 +5,7 @@
 import { Element } from './element';
 import { Font, FontInfo, FontStyle, FontWeight } from './font';
 import { Metrics } from './metrics';
-import { GroupAttributes, RenderContext, TextMeasure } from './rendercontext';
+import { RenderContext, TextMeasure } from './rendercontext';
 import { Tables } from './tables';
 import { normalizeAngle, prefix, RuntimeError } from './util';
 
@@ -96,6 +96,7 @@ export class SVGContext extends RenderContext {
 
     // Create a SVG element and add it to the container element.
     const svg = this.create('svg');
+    svg.setAttribute('pointer-events', 'none');
     this.element.appendChild(svg);
     this.svg = svg;
 
@@ -160,7 +161,7 @@ export class SVGContext extends RenderContext {
   }
 
   // Allow grouping elements in containers for interactivity.
-  openGroup(cls?: string, id?: string, attrs?: GroupAttributes): SVGGElement {
+  openGroup(cls?: string, id?: string): SVGGElement {
     const group = this.create('g');
     this.groups.push(group);
     this.parent.appendChild(group);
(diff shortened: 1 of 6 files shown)


src/svgcontext.ts
```diff
@@ -386,6 +384,12 @@ export class SVGContext extends RenderContext {
     return this;
   }
 
+  pointerRect(x: number, y: number, width: number, height: number): this {
+    const attributes = { opacity: '0', 'pointer-events': 'auto' };
+    this.rect(x, y, width, height, attributes);
+    return this;
+  }
+
```
ronyeh, line 392: So this solution draws an invisible rectangle on top of the glyph to capture events? Does that mean we are adding an extra shape for every note? Does this invisible rect cover the stems and flags and other parts, in addition to the note head?

I did verify that v4 mouseovers worked, while v5 glyphs have really large boxes that breaks the mouseovers (i.e., you can hover over empty space far above the glyph and it will still trigger the hover highlight). We might need to investigate to see if there is a better way.

rvilarl: Yes, it covers the box returned by StaveNote.getBoundingBox(). This is noteheads, stem and flag. No accidentals/ornaments/dots...

This is the only idea that I could think of to resolve the issue in  https://github.com/vexflow/vexflow/issues/155#issuecomment-1803319476

ronyeh: Yeah that's a tough one to solve. Ideally we only add these invisible rects if the developer wants to turn on interactivity. If a score is not interactive, we shouldn't need to add all these extra unused elements. 

Let me think about this for a bit. I can merge this weekend if I can't think of any better solution.

ronyeh:
I started looking at this PR again, and am currently doing some tests with the https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/pointer-events set to various values `visiblePainted | visibleFill | visibleStroke | visible`.

In the end, your solution is probably the only one that'll work, but I'll merge once I've tested the other options.

ronyeh:
Two thoughts:

1) I notice all our elements have a black fill AND a black stroke. Were the SMuFL fonts designed to include the stroke? Or were they designed such that they should **only be filled inside the SVG path**?

For example, our generated SVG looks like: 
```
<g class="vf-notehead" id="vf-auto2796" pointer-events="bounding-box" fill="black" stroke="black"><text stroke="black" font-size="30pt" x="153.56" y="95" fill="black"></text></g>
```

2) I'll go ahead and merge, but if we ever have a performance issue we should consider only creating a pointerRect when a user tries to add an event handler to a particular note group. That way we only have extra rectangles where needed.

PR #159, opened 2023-11-11
bravura only tests

rvilarl:
@ronyeh the tests are taking too long. I believe that at this stage we can test with bravura only.

### DIFF
--- a/tests/vexflow_test_helpers.ts
+++ b/tests/vexflow_test_helpers.ts
@@ -120,19 +120,7 @@ const SVG_TEST_CONFIG = {
   backend: Renderer.Backends.SVG,
   tagName: 'div',
   testType: 'SVG',
-  fontStacks: [
-    'Bravura',
-    'Finale Ash',
-    'Finale Broadway',
-    'Finale Maestro',
-    'Gonville',
-    'Gootville',
-    'Leland',
-    'Leipzig',
-    'MuseJazz',
-    'Petaluma',
-    'Sebastian',
-  ],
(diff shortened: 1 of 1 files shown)


ronyeh:
Instead of deleting, should we just comment out the other font names in that test file for now? Does this need to be a PR? Why can't you just comment out the extra fonts on your own local copy?

rvilarl:
Well there are complains running the tests. Jared thought that the machine froze. My old laptop has also problems.

tests/vexflow_test_helpers.ts
```diff
@@ -120,19 +120,7 @@ const SVG_TEST_CONFIG = {
...
-    'Finale Maestro',
-    'Gonville',
-    'Gootville',
-    'Leland',
-    'Leipzig',
-    'MuseJazz',
-    'Petaluma',
-    'Sebastian',
-  ],
+  fontStacks: ['Bravura'],
```
ronyeh, line 123: Let's just comment out all the other fonts for now. That will make it easier for developer to know which fonts they can test. (They can un-comment a font if they wish.)

Later on let's figure out how to allow people to select fonts via the UI.

ronyeh (changes requested):

tests/vexflow_test_helpers.ts
rvilarl: Hi Ron I decided to leave two so that the possible collaborators easily see that it is possible to use different fonts.

Issue #157, opened 2023-11-09
Dotted slurs are not available


rvilarl:
See measure 3

![image](https://github.com/vexflow/vexflow/assets/22865285/32963e19-1adf-4b11-bdcd-86c7e0c423cc)

rvilarl:
@mscuthbert are dotted slurs common?

mscuthbert:
They're pretty common (along with dotted ties) in some traditions and instruments. Timpani and snare rolls sometimes use them as do violin tremolos. Not all composers and presses use them but some do to mean "conceptually tied/slurred" even though in reality notes are being rearticulated

rvilarl:
> They're pretty common (along with dotted ties) in some traditions and instruments. Timpani and snare rolls sometimes use them as do violin tremolos. Not all composers and presses use them but some do to mean "conceptually tied/slurred" even though in reality notes are being rearticulated

Thanks! I will go for them then.

rvilarl:
@mscuthbert @AaronDavidNewman @sschmidTU @ronyeh @jaredjj3 @0xfe
The curves are now rendered using two lines this allows to specify a thickness. Is anyone using this?

![svg_Curve Simple_Curve Bravura svg_reference](https://github.com/vexflow/vexflow/assets/22865285/defe2896-2d39-46d6-8905-fffe252dba5c)

This implementation does not allow to use stroke-dasharray. 

Would it be fine to move to a single line implementation that would allow the usage of the dasharray?

![svg_Curve Simple_Curve Bravura svg_current](https://github.com/vexflow/vexflow/assets/22865285/21464ed0-b19f-4ab3-87a1-96a0a61cc296)

The line width above is 1, it could be increased.

0xfe:
The two lines allow for curves that taper at the ends, which is more consistent with how they're typically seen on sheets. Also, non-dotted curves are overwhelmingly the common case, so we shouldn't change our entire implementation to support it.

Instead, maybe create a separate LineCurve element for those corner cases? Or add an option to the existing Curve that uses a single line. (I prefer a separate element, but not a strong opinion.)

rvilarl:
Hi Mohit, I have to say that I liked more the single line :) . Here now with 2 points. I think that it also reduces the complexity.

![svg_Curve Simple_Curve Bravura svg_current](https://github.com/vexflow/vexflow/assets/22865285/d891c9db-1912-4a14-a02e-9dc2cf9d626b)

[...]

PR #160, opened 2023-11-11
linedash support

rvilarl:
intial step to support #157 

it also reesolves the deep copy in svgcontext.

no visual changes

### DIFF
--- a/src/svgcontext.ts
+++ b/src/svgcontext.ts
@@ -51,8 +51,6 @@ const TWO_PI = 2 * Math.PI;
 export interface State {
   state: Attributes;
   attributes: Attributes;
-  shadowAttributes: Attributes;
-  lineWidth: number;
 }
 
 /**
@@ -67,9 +65,7 @@ export class SVGContext extends RenderContext {
   height: number = 0;
   path: string;
   pen: { x: number; y: number };
-  lineWidth: number;
   attributes: Attributes;
-  shadowAttributes: Attributes;
   state: Attributes;
   stateStack: State[];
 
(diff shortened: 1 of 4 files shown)


rvilarl:
I have included the changes necessary to support dotted lines in curves. And modified two test to show the result. These are the only visual differences:

current
svg_Curve Rounded_Curve Bravura svg_current
reference
svg_Curve Rounded_Curve Bravura svg_reference
current
svg_Curve Simple_Curve Bravura svg_current
reference
svg_Curve Simple_Curve Bravura svg_reference

ronyeh:
The visual output looks good to me for now.

For reference, this is how Dorico does it:
https://steinberg.help/dorico/v1/en/dorico/topics/notation_reference/notation_reference_slurs_styles_r.html

I think that the ideal end result will be to have the dotted / dashed slur be thicker in the middle and taper at both ends (like the normal solid slur curve).

To do that, we might have to implement it as a filled shape, rather than a path with a dashed stroke.

But for now, we can do it this way.

src/svgcontext.ts
```diff
@@ -104,10 +100,9 @@ export class SVGContext extends RenderContext {
 
     this.path = '';
     this.pen = { x: NaN, y: NaN };
-    this.lineWidth = 1.0;
 
     const defaultFontAttributes = {
-      'font-family': Metrics.get('fontFamily'),
+      'font-family': Metrics.get('fontFamily') as string,
```
ronyeh, line 105: Curious why this cast was needed? Is it not normally a string?

ronyeh (approved):

src/svgcontext.ts
rvilarl: get returns any :(

PR #162, opened 2023-11-12
fix stavenote bounding box

rvilarl:
The topic ibounding box is not complete with this PR but I would like to check with a new alpha release if this goes in the good direction to resolve the issue #155 from @AaronDavidNewman. I have modified the stavenote bounding box test to include an accidental, an ornament and dots. This test is the only visual difference.

svg_StaveNote StaveNote_BoundingBoxes___Treble Bravura svg_current

### DIFF
--- a/src/element.ts
+++ b/src/element.ts
@@ -340,7 +340,7 @@ export class Element {
 
   /** Get the boundingBox. */
   getBoundingBox(): BoundingBox {
-    return new BoundingBox(0, -this.textMetrics.actualBoundingBoxAscent, this.width, this.height);
+    return new BoundingBox(this.x, this.y - this.textMetrics.actualBoundingBoxAscent, this.width, this.height);
   }
 
   /** Return the context, such as an SVGContext or CanvasContext object. */
--- a/src/stavenote.ts
+++ b/src/stavenote.ts
@@ -614,6 +614,9 @@ export class StaveNote extends StemmableNote {
       const bbFlag = this.flag.getBoundingBox();
       boundingBox.mergeWith(bbFlag.move(flagX, flagY));
     }
+    for (let i = 0; i < this.modifiers.length; i++) {
+      boundingBox.mergeWith(this.modifiers[i].getBoundingBox());
+    }
     return boundingBox;
   }
 
--- a/tests/stavenote_tests.ts
+++ b/tests/stavenote_tests.ts
@@ -27,6 +27,7 @@ import { Stem } from '../src/stem';
 import { StringNumber } from '../src/stringnumber';
 import { Stroke } from '../src/strokes';
 import { TickContext } from '../src/tickcontext';
+import { Ornament } from '../src';
 
 const StaveNoteTests = {
   Start(): void {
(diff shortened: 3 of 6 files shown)


rvilarl:
@ronyeh I placed the label because this is my priority to see if it fits to Aaron problem. :)

ronyeh:
Does this require a new alpha release? Just curious if Aaron can just check out this PR branch and see if it fixes the problem before we merge.

If that is not convenient, we could go ahead with this PR, but in general I prefer that we avoid making releases just to do a quick check, as it adds unnecessary commits to the repo, etc.

src/element.ts
```diff
@@ -340,7 +340,7 @@ export class Element {
 
   /** Get the boundingBox. */
   getBoundingBox(): BoundingBox {
-    return new BoundingBox(0, -this.textMetrics.actualBoundingBoxAscent, this.width, this.height);
+    return new BoundingBox(this.x, this.y - this.textMetrics.actualBoundingBoxAscent, this.width, this.height);
```
ronyeh, line 343: This looks like the core bug fix. Was the previous implementation just incorrect? Or are we changing the definition of the bounding box here?

src/stavenote.ts
```diff
@@ -614,6 +614,9 @@ export class StaveNote extends StemmableNote {
       const bbFlag = this.flag.getBoundingBox();
       boundingBox.mergeWith(bbFlag.move(flagX, flagY));
     }
+    for (let i = 0; i < this.modifiers.length; i++) {
+      boundingBox.mergeWith(this.modifiers[i].getBoundingBox());
```
ronyeh, line 618: OK, so we are defining the stavenote's bounding box to include its modifiers now. Will this break any other things that relied on the stavenote's bounding box being defined differently?

tests/stavenote_tests.ts
```diff
@@ -467,16 +468,16 @@ function drawBoundingBoxes(options: TestOptions, contextBuilder: ContextBuilder)
...
-      ctx,
-      (i + 1) * 25,
-      true /* drawBoundingBox */,
-      false /* addModifierContext */
-    );
+    notes.push(staveNote(noteStructs[i]));
+  }
+  notes[2].addModifier(new Ornament('tr'), 0);
+  notes[2].addModifier(new Accidental('b'), 0);
+  Dot.buildAndAttach([notes[2]], { all: true });
```
ronyeh, line 477: I always thought this API didn't fit the rest of the VexFlow API. I recall you and I settled on this due to problems with circular references.... is that correct? I wonder if we can improve the Dot.buildAndAttach() API in V5, and maybe make it more similar to the other ways of adding modifiers, e.g.:

```
notes[2].addModifier(new Dot())
```

Can we investigate this in a separate PR? I am curious if we no longer have the circular reference problem.

ronyeh: For example, the v3.0.9 API looked like:

```
note.addDot(index)
note.addDotToAll()
```

https://github.com/0xfe/vexflow/blob/00ec15c67ff333ea49f4d3defbd9e22374c03684/src/stavenote.js#L810-L823

It would be nice if we could restore that API, or at least make it look like `note.addModifier(new Dot())`

src/element.ts
rvilarl: It was incorrect, it was not taking into account the origin of the element.

src/stavenote.ts
rvilarl: No, there are no other visual differences.

tests/stavenote_tests.ts
rvilarl: note.addModifier(new Dot()) would not work for chords. You need one Dot per NoteHead.

ronyeh: OK, we can discuss this in a separate issue / PR. I think the older 3.0.9 API of addDot(index) or addDotToAll() was very friendly and easy to understand. If we can restore that API without running into any circular reference problems, I think we should.

ronyeh (approved):

PR #163, opened 2023-11-12
Element Style refactored: Bend

rvilarl:
Refactored to use the Element Style infrastructure. I changed all to black.

current
svg_Bend Bend_Phrase Bravura svg_current
reference
svg_Bend Bend_Phrase Bravura svg_reference

### DIFF
--- a/src/bend.ts
+++ b/src/bend.ts
@@ -59,10 +59,8 @@ export class Bend extends Modifier {
   protected phrase: BendPhrase[];
 
   public renderOptions: {
-    lineWidth: number;
     releaseWidth: number;
     bendWidth: number;
-    lineStyle: string;
   };
 
   /**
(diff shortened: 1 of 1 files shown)


src/bend.ts
```diff
@@ -101,11 +99,10 @@ export class Bend extends Modifier {
     this.xShift = 0;
     this.tap = '';
     this.renderOptions = {
-      lineWidth: 1.5,
-      lineStyle: '#777777',
       bendWidth: 8,
       releaseWidth: 8,
     };
+    this.setStyle({ lineWidth: 1.0, strokeStyle: 'black', fillStyle: 'black' });
```
ronyeh, line 105: Should this be customizable? I think the goal is to have all the rendering options in Metrics, and later allow developers to customize everything, right?

rvilarl: This is customizable calling SetSyle on the element. This is just the default set in the constructor.

There is a mix in the code where styles are in renderOptions. After Bend, I will be fixing the other ones.

ronyeh (approved):

PR #165, opened 2023-11-12
ornaments above stave

rvilarl:
There was no ornament test with low notes. Jared highlighted that they should stay about the Stave. 

I modified a test to lower the notes and Ornament.format to make the ornaments stay above.

Only one difference in the modified test

current
svg_Ornament Ornaments Bravura svg_current
reference
svg_Ornament Ornaments Bravura svg_reference
without the modification in Ornament, the ornaments go between lines.


ronyeh (approved):

rvilarl:
Thanks Ron!

PR #167, opened 2023-11-14
bounding box fixes

rvilarl:
I have started with tests and fixes for bounding boxes and I have found already a bug in Element.

Let us review and merge this group and follow on other PRs with the next derivates from Element.

No visual differences. Only new tests to check the bounding boxes.

### DIFF
--- a/src/annotation.ts
+++ b/src/annotation.ts
@@ -285,7 +285,9 @@ export class Annotation extends Modifier {
     }
 
     L('Rendering annotation: ', this.text, x, y);
-    this.renderText(ctx, x, y);
+    this.x = x;
+    this.y = y;
+    this.renderText(ctx, 0, 0);
     ctx.closeGroup();
     this.restoreStyle();
   }
--- a/tests/accidental_tests.ts
+++ b/tests/accidental_tests.ts
@@ -28,9 +28,9 @@ const AccidentalTests = {
     QUnit.module('Accidental');
     QUnit.test('Automatic Accidentals - Simple Tests', autoAccidentalWorking);
     const run = VexFlowTests.runTests;
-    run('Bounding Box', basic, { bbox: true });
+    run('Bounding Box', basic, { drawBoundingBox: true });
     run('Accidental Padding', formatAccidentalSpaces);
-    run('Basic', basic, { bbox: false });
+    run('Basic', basic, { drawBoundingBox: false });
     run('Stem Down', basicStemDown);
     run('Cautionary Accidental', cautionary);
     run('Accidental Arrangement Special Cases', specialCases);
@@ -283,7 +283,7 @@ function basic(options: TestOptions): void {
   f.draw();
 
   // Render bounding boxes
-  if (options.params.bbox === true) {
+  if (options.params.drawBoundingBox === true) {
     notes.forEach((note) => {
       const elements = note.getModifiersByType('Accidental');
       elements.forEach((element) => VexFlowTests.drawBoundingBox(f.getContext(), element));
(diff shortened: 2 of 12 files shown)


rvilarl:
New tests

canvas_Accidental Bounding_Box Bravura
canvas_Annotation Bounding_Box Bravura
canvas_Articulation Bounding_Box Bravura
canvas_ChordSymbol Bounding_Box Bravura
canvas_Clef Bounding_Box Bravura

src/annotation.ts
```diff
@@ -285,7 +285,9 @@ export class Annotation extends Modifier {
     }
 
     L('Rendering annotation: ', this.text, x, y);
-    this.renderText(ctx, x, y);
+    this.x = x;
+    this.y = y;
```
ronyeh, line 289: I see in the Element class there is a `setOrigin(x, y)`, but it looks surprisingly complicated. :-)

Do you know what the difference is between setOrigin(x,y) and what you're doing here?

There seems to be a x,y and a xShift,yShift and it is not clear what they each mean. We should clarify with comments in the Element class in a separate PR.

tests/accidental_tests.ts
```diff
@@ -281,6 +282,14 @@ function basic(options: TestOptions): void {
 
   f.draw();
 
+  // Render bounding boxes
+  if (options.params.bbox === true) {
```
ronyeh, line 286: Minor suggestion:

bbox could be `drawBoundingBox`. It doesn't matter if we are more verbose since these are just tests....

Then the code can be:

```
 if (options.params.drawBoundingBox) {
```

tests/articulation_tests.ts
```diff
@@ -240,6 +241,15 @@ function verticalPlacement(options: TestOptions, contextBuilder: ContextBuilder)
   ];
 
   Formatter.FormatAndDraw(ctx, stave, notes);
+
+  // Render bounding boxes
+  if (options.params.bbox === true) {
```
ronyeh, line 246: Same here. It could be written as:

 if (options.params.drawBoundingBox) {

tests/stavenote_tests.ts
```diff
@@ -9,6 +9,7 @@
 import { VexFlow } from '../src/vexflow';
 import { TestOptions, VexFlowTests } from './vexflow_test_helpers';
 
+import { Ornament } from '../src';
```
ronyeh, line 12: This works but should be written as:

```
import { Ornament } from '../src/ornament';
```

I know the original was already wrong, but we might as well fix it.

ronyeh (changes requested):
Just minor suggestions, see comments ⬆️.

src/annotation.ts
rvilarl: Yes it is really complicated and only used by Articulation. It took a while to get it without visual differences with the x/yShift. Normally x and y is related to position and xShift and yShift with formatting (distance between different elements)

rvilarl:
> Just minor suggestions, see comments ⬆️.

I am in Berlin without my linux computer. I will do the changes once I go back later this week. Thanks for the review!

tests/accidental_tests.ts
rvilarl: done

tests/articulation_tests.ts
rvilarl: done

tests/stavenote_tests.ts
rvilarl: done

ronyeh (approved):

PR #180, opened 2024-03-08
Fix Annotation and ChordSymbol justification padding

### LINKED ISSUE #178: Annotation width - Justification backwards
As I'm making bigger and bigger "annotations" with lyrics, I think that the computation of width is incorrect for LEFT or RIGHT justification:

```typescript
if (annotation.horizontalJustification === AnnotationHorizontalJustify.LEFT) {
    maxLeftGlyphWidth = Math.max(glyphWidth, maxLeftGlyphWidth);
    leftWidth = Math.max(leftWidth, textWidth) + Annotation.minAnnotationPadding;
} else if (annotation.horizontalJustification === AnnotationHorizontalJustify.RIGHT) {
    maxRightGlyphWidth = Math.max(glyphWidth, maxRightGlyphWidth);
    rightWidth = Math.max(rightWidth, textWidth);
}
```

this says that if we are left justified, that the amount of space to the left needs to be increased.  But really if we are left justified then the amount of space to the RIGHT needs to be increased.  And vice-versa.  So we can reverse the enums.

```typescript
if (annotation.horizontalJustification === AnnotationHorizontalJustify.RIGHT) {
    maxLeftGlyphWidth = Math.max(glyphWidth, maxLeftGlyphWidth);
    leftWidth = Math.max(leftWidth, textWidth) + Annotation.minAnnotationPadding;
} else if (annotation.horizontalJustification === AnnotationHorizontalJustify.LEFT) {
    maxRightGlyphWidth = Math.max(glyphWidth, maxRightGlyphWidth);
    rightWidth = Math.max(rightWidth, textWidth);
}
```

the default case of CENTER_STEM (or CENTER) is unchanged, which is why it escaped notice till now.

Similar code appears in ChordSymbol also, but I haven't tested it to see if it fixes it.

Testing with the longest one-syllable  English word.  Before:
Screenshot 2024-03-07 at 10 24 57

After:
Screenshot 2024-03-07 at 10 25 08

mscuthbert:
Fixes #178 

All the visual differences are things like this, where the first note was left justified, so it previously got extra space to the left, now it gets it to the right.

Reference:
canvas_Annotation TabNote_Annotations Bravura_reference

Current:
canvas_Annotation TabNote_Annotations Bravura_current

Reference:
canvas_ChordSymbol Top_Chord_Symbols Bravura_reference

Current:
canvas_ChordSymbol Top_Chord_Symbols Bravura_current

Added a new test with very long chord symbols and changed the lyrics test to start with a long lyric, left justified.

Reference:
svg_Annotation Lyrics Bravura svg_reference

Current:
svg_Annotation Lyrics Bravura svg_current

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,8 @@ ruby_sess.*
 # Jetbrains
 .idea/.gitignore
 .idea/VexFlow.iml
+.idea/vexflow5.iml
+.idea/copilot
 .idea/encodings.xml
 .idea/misc.xml
 .idea/modules.xml
--- a/src/annotation.ts
+++ b/src/annotation.ts
@@ -88,10 +88,10 @@ export class Annotation extends Modifier {
       const glyphWidth = note.getGlyphWidth();
       // Get the text width from the font metrics.
       const textWidth = annotation.getWidth();
-      if (annotation.horizontalJustification === AnnotationHorizontalJustify.LEFT) {
+      if (annotation.horizontalJustification === AnnotationHorizontalJustify.RIGHT) {
         maxLeftGlyphWidth = Math.max(glyphWidth, maxLeftGlyphWidth);
         leftWidth = Math.max(leftWidth, textWidth) + Annotation.minAnnotationPadding;
-      } else if (annotation.horizontalJustification === AnnotationHorizontalJustify.RIGHT) {
+      } else if (annotation.horizontalJustification === AnnotationHorizontalJustify.LEFT) {
         maxRightGlyphWidth = Math.max(glyphWidth, maxRightGlyphWidth);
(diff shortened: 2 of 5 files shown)


rvilarl:
Thanks

AaronDavidNewman:
I don't think this is the right fix - can't you just change it in the draw method, to be - instead of +?  Now we are mixing 'LEFT" and 'RIGHT'.

I think we are going to have to fix the test cases, but it can be in a different PR.  The issue is that there used to only be 3 justifications, but it wasn't clear whether they were vertical or horizontal - it did kind of a random mix.  When I updated to have both, I was so concerned about keeping the test results the same, to the point that I completely overlooked that I mixed up left and right.

Issue #182, opened 2024-03-08
Annotations -- should not affect ties


mscuthbert:
(on 4.3.2)

When computing the starting point of a tie there is an issue if an annotation like lyrics exist that are pretty long.  The tie gets smaller and smaller as in the last notes here.

<img width="315" alt="Screenshot 2024-03-07 at 16 53 42" src="https://github.com/vexflow/vexflow/assets/3521479/a941fa13-ddb2-40b3-a8b2-352a7961237d">

Eventually, if the lyric is long enough, the tie vanishes completely:

<img width="557" alt="Screenshot 2024-03-07 at 16 56 50" src="https://github.com/vexflow/vexflow/assets/3521479/d5ece4b8-65b8-442c-8641-fc0b9d903790">

mscuthbert:
Figured out a shim that works on this.  I'll happy contribute it into Vexflow 5, but I'd like comment on whether there's a danger to this approach which involves creating a new, disconnected modifierContext without lyrics, measuring it, and subtracting the difference.

```typescript
function getLyricWidthDifference(sn: StaveNote): {left: number, right: number} {
    const myModifierContext = sn.getModifierContext();
    myModifierContext.preFormat();  // just in case it hasn't been done yet.
    const myLeft = myModifierContext.getState().left_shift;
    const myRight = myModifierContext.getState().right_shift;

    const modifierContext2 = new ModifierContext();
    for (const cat of Object.keys((myModifierContext as any).members)) {
        if (cat === Annotation.CATEGORY) {
            continue;
        }
        for (const member of myModifierContext.getMembers(cat)) {
            modifierContext2.addMember(member);
        }
    }
    modifierContext2.preFormat();
    const noLyricLeft = modifierContext2.getState().left_shift;
    const noLyricRight = modifierContext2.getState().right_shift;
    return {
        left: myLeft - noLyricLeft,
        right: myRight - noLyricRight,
    };
}

const original_StaveNote_getTieRightX = StaveNote.prototype.getTieRightX;
StaveNote.prototype.getTieRightX = function getTieRightX(): number {
    const originalRight = original_StaveNote_getTieRightX.bind(this)();
    return originalRight - getLyricWidthDifference(this).right;
};
```

AaronDavidNewman:
Wouldn't you have the same problem with modifiers that come after the note head, like ornaments?  It seems like you would.

[...]

PR #183, opened 2024-03-08
StaveTie -- move rendering logic out of draw

mscuthbert:
Moving the rendering logic (where to start, where to end, etc.) out of the draw() method for StaveTie so that it can be tested separately from the visual tests.  (And hopefully expanded soon).

Added ability to configure the cutoff point for where a tie gets classified as a short tie and what the effects of that are.

Added a visual test (w/ some application logic tests) of a more standard use of ties: two chords that have the same pitches, and with two different ties between the same chords representing different directions.  (Surprisingly, this most standard use of ties was not being tested).
Screenshot 2024-03-08 at 12 21 20

Still learning the factory system.  :-)

Let me know if changes like this are appreciated, neutral, or -1.  

Preparatory work for fixing the ties w/ lyrics problem in #182

### DIFF
--- a/tests/annotation_tests.ts
+++ b/tests/annotation_tests.ts
@@ -7,31 +7,27 @@
 //       Did a previous version of the API accept a number as the fourth argument?
 //       We removed the fourth argument from all of our test cases.
 
-import {VexFlow} from '../src/vexflow';
-import {TestOptions, VexFlowTests} from './vexflow_test_helpers';
+import { VexFlow } from '../src/vexflow';
+import { TestOptions, VexFlowTests } from './vexflow_test_helpers';
 
-import {
-  Annotation,
-  AnnotationHorizontalJustify,
-  AnnotationVerticalJustify
-} from '../src/annotation';
-import {Articulation} from '../src/articulation';
-import {Beam} from '../src/beam';
-import {Bend} from '../src/bend';
-import {ElementStyle} from '../src/element';
-import {Formatter} from '../src/formatter';
-import {Metrics} from '../src/metrics';
-import {ModifierPosition} from '../src/modifier';
-import {Registry} from '../src/registry';
-import {ContextBuilder} from '../src/renderer';
-import {Stave} from '../src/stave';
-import {StaveNote, StaveNoteStruct} from '../src/stavenote';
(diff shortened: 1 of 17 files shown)


rvilarl:
Changes like that are appreciated from my side. @ronyeh what do you think?

ronyeh:
Haven't had a chance to review this yet, but just wanted to say that I appreciate any and all contributions to improving code & documention quality.

I like that you are adding comments as you go. It'll definitely help everyone who uses VexFlow (including ourselves)!

mscuthbert:
Great!  Then I'll continue to add comments as I investigate.  Sometimes the comments might be in weird places -- for instance, if I'm looking to see if Beams affect formatter to lead to poor behavior, and I conclude that it didn't, but I had to figure out something about beaming along the way I'll leave any comments I made there.

ronyeh:
Even if comments are fuzzy or incomplete, or even if they are of the "Huh?" type, I think it will help. When we all can admit to not understanding a part of VexFlow, we can work together to improve the code clarity & quality (or even fix any subtle bugs that confusing code can introduce).

tests/annotation_tests.ts
```diff
@@ -7,31 +7,27 @@
...
+import { Registry } from '../src/registry';
+import { ContextBuilder } from '../src/renderer';
+import { Stave } from '../src/stave';
+import { StaveNote, StaveNoteStruct } from '../src/stavenote';
+import { Stem } from '../src/stem';
+import { TabNote, TabNoteStruct } from '../src/tabnote';
+import { TabStave } from '../src/tabstave';
+import { Tickable } from '../src/tickable';
+import { Vibrato } from '../src/vibrato';
+import { Voice } from '../src/voice';
```
ronyeh, line 30: My copy of annotation_tests.ts already has these eslint fixes.

Can you fetch the most recent revision and rebase off of main? Maybe these changes will disappear?

ronyeh: https://github.com/vexflow/vexflow/blob/main/tests/annotation_tests.ts

mscuthbert: Changes disappeared with the last real commit and pulling "main".

PR #188, opened 2024-03-10
Formatter: mean calculation + pull out algorithms

### LINKED ISSUE #187: Bug in Mean calculation in formatter?
There seems to be a bug in Formatter.tune()'s embedded function "updateStats".

It is called for each Tickable with its duration (converted to a string of a reduced Fraction, like '1/2') and the amount of space off from the ideal for that duration it has:

```typescript
    function updateStats(duration: string, space: number) {
      const stats = durationStats[duration];
      if (stats === undefined) {
        durationStats[duration] = { mean: space, count: 1 };
      } else {
        stats.count += 1;
        stats.mean = (stats.mean + space) / 2;
      }
    }
```

The Mean seems to be calculated incorrectly.  If there are 3 quarter notes with space-differences of -2, 0, 2, then their mean deviation is obviously 0.  But here it would do:

```
-2  => mean = -2
0  => mean = -1
2 => mean = 0.5
```

I have a patch for this along with some other light refactors (putting the embedded function "move" on TickContext rather than deeply nested inside Formatter, for instance) I'm happy to PR if this sounds like a solution.  Just wanted to make sure that this wasn't intentional.  Note though that this correction is making no difference right now in the generated images (except for the annotations at the bottom) so I'm not entirely sure why yet.

There's a second point I'm not sure is intentional or not:  TickContexts with more Tickables potentially get a lot more weight in the tuning process (because the deviation cost is calculated as a sumArray, not a sumArray / numTickables):

```typescript
const cost = -sumArray(context.getTickables().map((t) => t.getFormatterMetrics().space.deviation));
```
rather than

```typescript
const cost = -sumArray(context.getTickables().map((t) => t.getFormatterMetrics().space.deviation)) 
    / context.getTickables().length;  // with div by 0 checks...
```

Also happy to include that fix, but maybe we care more about proper positioning the more tickables there are in a context?  (seems odd though).

mscuthbert:
Fixes #187 and implements part of the philosophy of #183.

Fixes the calculation of means by adding a "total" to `Formatter().durationStats` so that mean deviation is calculated properly.

`Formatter().tune()` previously embedded a powerful `move(tickContext, shift)` routine inside it that did not use any scope or `this` variables.  Rewrote as `TickContext().move(shift)`

`Formatter().tune()` calculated the cost of space deviations as `-sumArray(context.getTickables().map((t) => t.getFormatterMetrics().space.deviation));` which does not use any scope or `this` variables.  Moved to `TickContext().getDeviationCost()` (without the negative sign).  Added `TickContext().getAverageDeviationCost()` which I believe would fix a bug that contexts with more tickables are weighed more heavily, but did not use this calculation yet, pending discussion.

Docs and typo fixes:

`Gruntfile.js` -- noted that the standard developer build system will wipe out reference images -- gave an alternative that preserves them.

`FormatParams.alignRests` (and other alignRests places) -- add notes that this aligns rests vertically not horizontally.

Rename some internal `tick` variables to `tickContext` -- I'd like us to move to using `tick` to mean a `Fraction` object consistently.

Added warning to `Fraction.LCMM` -- it actually seems like a bug to me, but didn't want to change w/o discussion.

Generated images -- only changes are to the values reported for space deviations in Factory classes, which abs(lower) the mean differences in most cases.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -19,6 +19,10 @@ DEVELOP
 
 grunt watch
   - FAST: Watch for changes and produces the debug CJS libraries in build/cjs/.
+  - Deletes the reference images and anything else in build first.
+
+grunt webpack:watchDebug
+  - Same as grunt watch without deleting the build directory first.
 
 grunt watch:esm
   - FAST: Watch for changes and build the ESM libraries in build/esm/.
--- a/src/formatter.ts
+++ b/src/formatter.ts
@@ -147,7 +147,7 @@ function getRestLineForNextNoteGroup(
     }
   }
 
-  // Locate the mid point between two lines.
+  // Locate the midpoint between two lines.
   if (compare && currRestLine !== nextRestLine) {
     const top = Math.max(currRestLine, nextRestLine);
     const bot = Math.min(currRestLine, nextRestLine);
@@ -1006,7 +1006,7 @@ export class Formatter {
       context.move(shift, prevContext, nextContext);
 
       // Q(msac): Should the cost by normalized by the number
-      // of tickables at this position?
+      // of tickables at this position?  If so, switch this to getAverageDeviationCost()
       const cost = -context.getDeviationCost();
       if (cost > 0) {
         shift = -Math.min(context.getFormatterMetrics().freedom.right, Math.abs(cost));
--- a/src/formatter.ts
+++ b/src/formatter.ts
@@ -39,7 +39,7 @@ export interface FormatterOptions {
 }
 
 export interface FormatParams {
-  alignRests?: boolean;
+  alignRests?: boolean; // align rests vertically with neighboring notes.
   stave?: Stave;
   context?: RenderContext;
   autoBeam?: boolean;
@@ -189,7 +189,7 @@ export class Formatter {
   protected modifierContexts: AlignmentModifierContexts[];
   protected voices: Voice[];
   protected lossHistory: number[];
-  protected durationStats: Record<string, { mean: number; count: number }>;
+  protected durationStats: Record<string, { mean: number; count: number; total: number }>;
 
   /**
    * Helper function to layout "notes" one after the other without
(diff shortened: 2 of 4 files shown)


AaronDavidNewman:
The stats change is correct.  However, it doesn't look like anything looks at the return value of `formatter.evaluate()`.  Could we just...not call it?  It might save a couple of clicks.

I have no idea what `formatter.tune` does.  Do you?   There is only one test case for it, and it does not make it looks better.

mscuthbert:
> The stats change is correct. However, it doesn't look like anything looks at the return value of `formatter.evaluate()`. Could we just...not call it? It might save a couple of clicks.

There are a ton of side effects in `evaluate()`:  it sets `this.contextGaps` and changes the `formatterMetrics.freedom` for each of the TickContexts and Notes, and stores its value in `this.totalCosts` and updates `this.lossHistory`.  

`evaluate()` is called by both tune() and preFormat().  (Where renderingContext is never used...)
 
> I have no idea what `formatter.tune` does. Do you? There is only one test case for it, and it does not make it looks better.

Nope -- I discovered this while I was trying to figure out `tune()` but then I realized it was only being called from Factory, and when I called it myself on my own formatter, it just made things worse.  So much worse, in fact, that I think there might be a sign error (possibly on  the cost function).  

Most of these PRs right now are just things I'm discovering while trying to figure out what Formatter actually does, where, and why.  I'll try to keep adding comments along the way, because if I don't know how it works, and you don't know how it works, I'll wonder if there's any single person on the project who actually knows how the whole thing works.

AaronDavidNewman:
> There are a ton of side effects in evaluate(): it sets this.contextGaps and changes the formatterMetrics.freedom for each of the TickContexts and Notes, and stores its value in this.totalCosts and updates this.lossHistory

It looks like contextGaps are used in the formatter unit test cases to make those pretty graphs.   So maybe we could make this call optional, someday.  Since it is only called at the end of format, there's no use in computing the stats in most applications, they just get lost when the formatter goes out of scope.

> when I called it myself on my own formatter, it just made things worse

right.  I suspect 'tune' and 'evaluate' were meant to be used together to learn parameters for the formatter.  Interesting idea, but no scrolls exist...

mscuthbert:
no scrolls exist?

mscuthbert:
Discussion has been fruitful but about parts of the system that existed before the proposed changes.  Bug-fix is confirmed.  Merging -- more works still needs to be done.

PR #189, opened 2024-03-10
Annotation - remove outdated comment

mscuthbert:
Minor.

A change to v5 format in  #51 left a comment that is now outdated.  Annotation no longer changes anything in RenderContext except through applyStyle(), so there is no need for an additional ctx.save() or ctx.restore().

https://github.com/vexflow/vexflow/commit/d6c878989e3e7f9723d7fcf63fb039faef7c66df

Fix param typing in setJustification

### DIFF
--- a/src/annotation.ts
+++ b/src/annotation.ts
@@ -224,10 +224,6 @@ export class Annotation extends Modifier {
     const start = note.getModifierStartXY(ModifierPosition.ABOVE, this.index);
 
     this.setRendered();
-
-    // Apply style might not save context, if this.style is undefined, so we
-    // still need to save context state just before this, since we will be
-    // changing ctx parameters below.
     this.applyStyle();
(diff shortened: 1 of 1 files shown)


mscuthbert:
A Question for the group -- should `applyStyle()` and `removeStyle()` go ahead and do a `ctx.save()`/`ctx.restore()` if there is a context defined whether or not style is defined?  It seems like the small overhead of keeping an extra state-push/pop is worth it for avoiding the subtle bugs that occur when state isn't fully restored.

AaronDavidNewman:
> should applyStyle() and removeStyle() go ahead and do a ctx.save()/ctx.restore()

Yes, I agree.

mscuthbert:
Okay, will make an issue to make applyStyle/removeStyle always set a stack trace, and merging as no objections to the removal of the comment raised.

PR #190, opened 2024-03-10
Remove unneeded test on StaveModifier

mscuthbert:
A test that each StaveModifier has a draw() method goes back to the pre-ES6 class version of Vexflow and is unnecessary now.

First added way back in 2012 as:
https://github.com/vexflow/vexflow/commit/f5e562519dffd0e5931c136218ab36bbc2cf1cb3

Added a type to remind ourselves that these are StaveModifiers not Modifiers.

Q: Stave().getModifiers returns a different array than Stave.modifiers if position or category is specified, but returns the original array if neither is specified.  I wonder if line 567 should be changed to return [...this.modifiers] so that in all cases `.modifiers` remains immutable?

### DIFF
--- a/src/stave.ts
+++ b/src/stave.ts
@@ -564,7 +564,7 @@ export class Stave extends Element {
     const noPosition = position === undefined;
     const noCategory = category === undefined;
     if (noPosition && noCategory) {
-      return this.modifiers;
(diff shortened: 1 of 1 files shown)


mscuthbert:
Merging as uncontroversial, no objections raised.

PR #193, opened 2024-03-11
Remove hash from private properties

### LINKED ISSUE #179: Avoid using # for private members.
Hi @mscuthbert,

Thanks for your feedback. I agree with your reasoning. It does make it easier to debug and/or extend (if you really really know what you're doing). Have a private keyword is probably a good enough way to indicate to developers that we don't intend for a property or method to be publicly consumed. 

We were probably too aggressive with "using the new shiny things" in JavaScript. I can look into reverting this.

> Ohh... I had missed this until now.  Can I register a strong -1 against this change?  VexFlow is not a super mature stable system, nor is it a proprietary system with backend integration that must be closely protected against people accidentally screwing up their accounts, or any of the other major reasons for marking things private.
> 
> With the older "private x: string = ..." the system is telling a user "Please don't touch this.  You've been warned".  With the `#` system, a modern browser makes it impossible to access the information -- it's just gone.
> 
> This means: no writing debugging code to figure out what the value of a private variable is.  No setting `(element as any).style.stroke = ` to do something that Vexflow didn't originally intend to have public.  I think that the `#` tag is too powerful and limiting to recommend.
> 

_Originally posted by @mscuthbert in https://github.com/vexflow/vexflow/issues/10#issuecomment-1984444055_

mscuthbert:
For many of them I looked up in the past whether they were protected or private (or if they were mentioned in docs) and set protected or private accordingly.  For the rest I used my best judgment about whether I would ever want to query them outside of the system or not.

Fixes #179 

Add runtime tests for two cases where zero-width element would cause infinite loops -- these came up when my code was buggy, but figured it's worth keeping the tests in.

Running visual differences now.

### DIFF
--- a/src/beam.ts
+++ b/src/beam.ts
@@ -68,26 +68,27 @@ export class Beam extends Element {
   postFormatted: boolean;
   slope: number = 0;
 
-  readonly #stemDirection: number;
-  readonly #ticks: number;
+  private readonly _stemDirection: number;
+  private readonly _ticks: number;
 
-  #yShift: number = 0;
-  #breakOnIndexes: number[];
-  #beamCount: number;
-  #unbeamable?: boolean;
+  protected yShift: number = 0;
+  private breakOnIndexes: number[];
+  private _beamCount: number;
+  // note that this is never set and is a private property.  Remove?
+  private unbeamable?: boolean;
 
   /**
    * Overrides to default beam directions for secondary-level beams that do not
    * connect to any other note. See further explanation at
    * `setPartialBeamSideAt`
    */
-  #forcedPartialDirections: {
+  private forcedPartialDirections: {
     [noteIndex: number]: PartialBeamDirection;
   } = {};
 
   /** Get the direction of the beam */
   getStemDirection(): number {
-    return this.#stemDirection;
+    return this._stemDirection;
   }
 
   /**
@@ -459,18 +460,19 @@ export class Beam extends Element {
     }
 
     // Validate beam line, direction and ticks.
-    this.#ticks = notes[0].getIntrinsicTicks();
+    this._ticks = notes[0].getIntrinsicTicks();
 
-    if (this.#ticks >= Tables.durationToTicks('4')) {
+    if (this._ticks >= Tables.durationToTicks('4')) {
+      // Q(MSAC) -- what about beamed half-note measured tremolos?
       throw new RuntimeError('BadArguments', 'Beams can only be applied to notes shorter than a quarter note.');
     }
 
     let i; // shared iterator
     let note;
 
-    this.#stemDirection = notes[0].getStemDirection();
+    this._stemDirection = notes[0].getStemDirection();
 
-    let stemDirection = this.#stemDirection;
+    let stemDirection = this._stemDirection;
     // Figure out optimal stem direction based on given notes
 
     if (autoStem && isStaveNote(notes[0])) {
@@ -486,15 +488,15 @@ export class Beam extends Element {
       note = notes[i];
       if (autoStem) {
         note.setStemDirection(stemDirection);
-        this.#stemDirection = stemDirection;
+        this._stemDirection = stemDirection;
       }
       note.setBeam(this);
     }
 
     this.postFormatted = false;
     this.notes = notes;
-    this.#beamCount = this.getBeamCount();
-    this.#breakOnIndexes = [];
+    this._beamCount = this.getBeamCount();
+    this.breakOnIndexes = [];
     this.renderOptions = {
       beamWidth: 5,
       maxSlope: 0.25,
@@ -525,7 +527,7 @@ export class Beam extends Element {
 
   /** Set which note `indexes` to break the secondary beam at. */
   breakSecondaryAt(indexes: number[]): this {
-    this.#breakOnIndexes = indexes;
+    this.breakOnIndexes = indexes;
     return this;
   }
 
@@ -543,7 +545,7 @@ export class Beam extends Element {
    * ```
    */
   setPartialBeamSideAt(noteIndex: number, side: PartialBeamDirection) {
-    this.#forcedPartialDirections[noteIndex] = side;
+    this.forcedPartialDirections[noteIndex] = side;
     return this;
   }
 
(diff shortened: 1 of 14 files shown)


mscuthbert:
No reference images changed.

src/beam.ts
```diff
@@ -68,26 +68,27 @@ export class Beam extends Element {
...
+  private readonly _stemDirection: number;
+  private readonly _ticks: number;
 
-  #yShift: number = 0;
-  #breakOnIndexes: number[];
-  #beamCount: number;
-  #unbeamable?: boolean;
+  protected yShift: number = 0;
+  private breakOnIndexes: number[];
+  private _beamCount: number;
```
rvilarl, line 76: what is the meaning of the underscore?

mscuthbert: If there was an underscore before I restored it. Mostly when there was a getter or setter for the property that conflicted (like text and width). Since the PR that put these in wasn't cleanly resolved it was hard to tell what the case was. This looked like a prime case where the attribute is likely to be replaced by a property that computes beam counts on the fly. But I'll happily replace it back if it seems confusing. But isn't beamCount also a method?

ronyeh:
> For many of them I looked up in the past whether they were protected or private

To be clear, nothing was `private` or `protected` in VexFlow 3.0.9. These labels were introduced when Rodrigo and I moved the code base over to TypeScript. We might have had a good reason at that time, or we might have done it without thinking too deeply about it. :-D

I'd say we should choose whatever makes the most sense **today**. And in general I'd tend to favor protected unless it is an obviously internal value for temporary calculations that no one (even subclasses) should be able to access.

Here is some prior discussion:

https://github.com/0xfe/vexflow/issues/1082

https://github.com/0xfe/vexflow/issues/1082

* * * 

My honest opinion is that as a library targeted towards developers who like to extend and add features, we should make the library as "hackable" as possible. 

So I'd be OK with actually converting these access modifiers to **comments** and things like underscore prefixes.

For example:

```ts

/* private */ __internalTemporaryCacheValueForFormatting: number = 0; // We use this to adjust formatting of staves. You do not need to access this value, but feel free to do so for testing / debugging.

/* protected */ _yShift: number = 0; // Please use the getYShift() / setYShift(val) accessors. However, subclasses may edit this value directly if you really need to.

```

What do you all think about something like this?

This would make the VexFlow library as hackable as possible. We can let developers know if they access any fields prefixed with a double underscore, they're really on their own. We can feel free to change the variable name whenever we feel like it.

Variables prefixed with single underscores are part of the official API, but should really only be accessed by subclasses.

Thoughts?

ronyeh:
Back in 2021 we were discussing whether to make render_options protected....

I made a comment at that time that before the TS migration, VexFlow was a pure JS project and everything was effectively public. (See below.)

I still think that it's OK if everything is publicly accessible, as long as we have clear comments as to if / when / how a field should be accessed. I'm up for using underscores to discourage public access if you all are OK with that.

It would certainly make testing and debugging easier.


> > please note that protected has only been added to 4 classes (Beam, Note, Stavetie & TextBracket), in the other 13 classes, this attribute was already protected.
> 
> @rvilarl we need to remind ourselves that before March 2021, when you and I started working on this project, all member variables were effectively "public" even if Mohit didn't intend them to be. We probably introduced the `protected` labels during the TS migration. So we should consider how our 4.0 release will affect users of the project. I'm fine with introducing breaking changes (and we have done quite a few), but we need to accommodate the developers who use this project, like @mscuthbert and @sschmidTU and others.
> 
> Anyways, we should get more feedback, including from @0xfe.
> 
> Questions
> * Should we make render_options protected and then provide individual accessors to each parameter?
> * Should we eliminate render_options completely, and assign the validated parameters directly to instance variables (while providing the necessary accessors)?
> * Something else?
> 

_Originally posted by @ronyeh in https://github.com/0xfe/vexflow/issues/1070#issuecomment-942762522_

rvilarl:
I would rather stay with the standard protected / private keywords. Javascript provides hacks to access a private member M of class C: 
```javascript
(C as any).M
```
Reducing scope avoids unintended use.

mscuthbert:
I'm with @rvilarl on using the Typescript standard to guard private and protected variables.  But overall, I think that there's some overuse of all private/protected.  I'd like to see them used only when we're sure that there's another way to access the same functionality.  For instance, make `.width` protected when there is a `getWidth()` that also checks that formatting has been run first.

In general I don't see much need for private/protected methods (but I definitely prefer them to inner-wrapped functions which are unable to be queried from the outside).

I could also see more of an argument for private/protected in the "Factory" module, since that is supposed to be an "easy to use/safe" way of using Vexflow.  But the rest of the system is a low-level library for building new things, so let's encourage people to build new things with it!

mscuthbert:
Is this good to merge -- at least for now -- to get the private properties at least hackable by downstream users again?  Thanks!

ronyeh:
Sounds good to me! If we figure out any good ways to avoid the leading
underscore for some of those fields, that’d be awesome too.

tests/flow-headless-browser.html
```diff
@@ -48,6 +48,8 @@
     <script>
       QUnit.config.autostart = false;
       QUnit.config.noglobals = true;
+      // uncomment below to detect infinite loops in tests.
+      // QUnit.testStart(d => console.log(d.name, d.module));
```
ronyeh, line 52: Cool idea!

src/vibrato.ts
```diff
@@ -68,8 +69,12 @@ export class Vibrato extends Modifier {
   setVibratoWidth(width: number): this {
     this.renderOptions.width = width;
     this.text = String.fromCodePoint(this.renderOptions.code);
+    const myWidth = this.getWidth();
+    if (!myWidth) {
```
ronyeh, line 73: I'm gonna go ahead and merge this PR, but in the future I think we should test for zero with === 0 instead of !.

if (!boo) should probably be reserved for boolean variables, and at least I tend to read them as booleans at first glance.

ronyeh (approved):

src/vibrato.ts
sschmidTU: It also checks for undefined/null.
But yes, if undefined isn't possible it basically checks === 0.
If the idea is that the width should be positive, I would check >= 0, because we probably also don't want negative width.

PR #199, opened 2024-04-25
Use string literal union for key signature type

judocode:
Key signatures are a known set of string literals, so this improves the typing on them

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -297,7 +297,7 @@ export class Factory {
     return timeSigNote;
   }
 
-  KeySigNote(params: { key: KeySpec; cancelKey?: string; alterKey?: string[] }): KeySigNote {
+  KeySigNote(params: { key: KeySpec; cancelKey?: KeySpec; alterKey?: string[] }): KeySigNote {
     const keySigNote = new KeySigNote(params.key, params.cancelKey, params.alterKey);
     if (this.stave) keySigNote.setStave(this.stave);
     keySigNote.setContext(this.context);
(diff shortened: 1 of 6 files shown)


rvilarl:
@judocode Thanks! I like the PR and the approach. There are missing changes if we go this way: CancelKey should also be KeySpec and there are a number of functions in Tables that would no longer be needed. Before making the change, let us ask.
@ronyeh @AaronDavidNewman @mscuthbert @jaredjj3 what do you think?

judocode:
Thanks @rvilarl ! I updated to add KeySpec to cancelKey. Should I do the same for alterKey?

rvilarl:
No this is an array of accidentals.

mscuthbert:
My feeling is that it's generally best to add the union string type at least to keep things compatible.  It can be annoying for a developer to have had code that works something like this:

```
const tonic: KeySpec = thisPitch.name;  // C, C#, etc.
if (mode == 'major') {
     return new KeySignature(tonic)
} else {
     return new KeySignature(tonic + 'm')
}
```

and then suddenly long existing code won't compile because `tonic + 'm'` is not a KeySpec.  I only find string literal types to be helpful when there are very few (like 2 or 3) and they're not things likely to be constructed from other strings.  Things like this, I think are better considered Runtime Errors than needing typing.

mscuthbert:
@judocode -- I really appreciate your contribution I should have added. This project needs new contributors with the energy and thought you brought to this PR. Even though I had some disagreements with it as a particular, I hope to keep seeing your contributions going forward. :-)

PR #201, opened 2024-05-02
Use of nullish coalescing operator with minus

eliot-akira:
I was building a new project that uses Vexflow, and saw a warning from ESBuild.

> [WARNING] The "??" operator here will always return the left operand [suspicious-nullish-coalescing]

```
vexflow/src/stave.ts:659:22:
    659 │         widths.left = -layoutMetrics.xMin ?? 0;
        ╵                       ~~~~~~~~~~~~~~~~~~~
```

>  The left operand of the "??" operator here will never be null or undefined, so it will always be returned. This usually indicates a bug in your code.

The underlined part of the code has a minus `-` that always tries to coerce `layoutMetrics.xMin` into a number; when it's undefined that results in `NaN`, so it never reaches `0`.

To correct this I changed it to:

```js
widths.left = -(layoutMetrics.xMin ?? 0);
```

The warning no longer appears. This seems to be the intended behavior, but maybe that minus is not supposed to be there at all? Later in the code, `widths.left` is used like:

```js
x -= widths.left;
```

### DIFF
--- a/src/stave.ts
+++ b/src/stave.ts
@@ -656,7 +656,7 @@ export class Stave extends Element {
           widths.right = layoutMetrics.xMax ?? 0;
           widths.paddingRight = layoutMetrics.paddingRight ?? 0;
         }
-        widths.left = -layoutMetrics.xMin ?? 0;
+        widths.left = -(layoutMetrics.xMin ?? 0);
         widths.paddingLeft = layoutMetrics.paddingLeft ?? 0;
 
         if (i === endModifiers.length - 1) {


mscuthbert:
Absolutely the correct solution.  Thanks.

The second part -- changing the sign of left -- let's wait on.  I think there are enough moving parts that changing a value like this in the middle of a V5 work might make life harder for some devs.

ronyeh:
Thanks for the catch, @eliot-akira !

eliot-akira:
Glad to contribute a tiny bit. :) I've been fascinated by Vexflow for some years, excited for this new version 5. I'd love to contribute/participate more when I find the time and opportunity.

A thing I'm building with Vexflow is called [mooz.app](https://mooz.app/scores/), a work in progress, where I used [OpenSheetMusicDisplay](https://github.com/opensheetmusicdisplay/opensheetmusicdisplay) that bundles a fork of Vexflow 1.2.93. I started working on replacing the notation feature with Vexflow 5 and [VexML](https://github.com/stringsync/vexml/). As a long-term ambitious goal, I'm hoping to build a "score player" with audio instruments synced to a cursor (or highlight per measure), as a practical tool for learning/teaching music.

PR #212, opened 2024-06-29
fix unison notehead accidental collision with stems down

### LINKED ISSUE #210: Unison notehead accidental collision with stems down
When stems go down, it appears that two-notehead augmented unisons have overlap between accidental and the left notehead.  

Report from Joseph VanderStel

```typescript
const note_obj = new VF.StaveNote({
    "keys":["Bn/5","B#/5"],
    "duration":"h",
    "stemDirection":-1,
});
note_obj.addModifier(new VF.Accidental('n'), 0);
note_obj.addModifier(new VF.Accidental('#'), 1);
```
Screenshot 2024-06-27 at 00 54 36

rvilarl:
fixes #210 

Minor code fix.

Accidental padding test modified to show fix
image

### DIFF
--- a/src/stavenote.ts
+++ b/src/stavenote.ts
@@ -569,7 +569,7 @@ export class StaveNote extends StemmableNote {
       if (lastLine === undefined) {
         lastLine = line;
       } else {
-        if (Math.abs(lastLine - line) === 0.5) {
+        if (Math.abs(lastLine - line) < 1) {
           this.displaced = true;
           props.displaced = true;
 
--- a/tests/accidental_tests.ts
+++ b/tests/accidental_tests.ts
@@ -163,9 +163,9 @@ function formatAccidentalSpaces(options: TestOptions): void {
       duration: '8d',
     }).addModifier(new Accidental('##'), 0),
     new StaveNote({
-      keys: ['b/4'],
+      keys: ['Bb/4','Bn/4'],
       duration: '16',
-    }).addModifier(new Accidental('b'), 0),
+    }).addModifier(new Accidental('b'), 0).addModifier(new Accidental('n'), 0),
     new StaveNote({
       keys: ['f/3'],
       duration: '8',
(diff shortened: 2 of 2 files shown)


tests/accidental_tests.ts
```diff
@@ -163,9 +163,9 @@ function formatAccidentalSpaces(options: TestOptions): void {
       duration: '8d',
     }).addModifier(new Accidental('##'), 0),
     new StaveNote({
-      keys: ['b/4'],
+      keys: ['Bb/4','Bn/4'],
       duration: '16',
-    }).addModifier(new Accidental('b'), 0),
+    }).addModifier(new Accidental('b'), 0).addModifier(new Accidental('n'), 0),
```
mscuthbert, line 168: This is great!  Thank you!  I wonder if we should start as a good habit, adding in comments that explain the purpose of each test so that in the future if they're refactored, people know what is happening.  For instance, just adding here:

```javascript
\\ test that unisons with different accidentals layout properly with stems down
```

and later for D/4:

```javascript
\\ test that unisons with different accidentals layout properly with stems up
```

this way anyone who comes along later knows that replacing `Bb/4, Bn/4` with `Ab/4, An/4` isn't an equivalent test.

src/stavenote.ts
```diff
@@ -569,7 +569,7 @@ export class StaveNote extends StemmableNote {
       if (lastLine === undefined) {
         lastLine = line;
       } else {
-        if (Math.abs(lastLine - line) === 0.5) {
+        if (Math.abs(lastLine - line) < 1) {
```
mscuthbert, line 572: Thanks for finding this.  wow.  Maybe good to add a comment like :`< 1 means == 0.5 (second) or 0.0 (unison)`

I totally see how this would fix the problem.  I just have no clue why it worked in the first place for stems up!

mscuthbert (approved):
Wow!  Looks amazing to me.  Thank you so much for taking the time to find this obscure bug that I didn't know where it was!    I put in a couple of comments that, if you wanted to take, would be nice to have in a future PR to save everyone the trouble.

tests/accidental_tests.ts
mscuthbert: (oops, I guess it is since there's a beamed note.  :-)

src/stavenote.ts
rvilarl: With stems up the notehead is always displaced to the left

PR #214, opened 2024-07-11
fix clear, no scale required

### LINKED ISSUE #202: Scaling SVG doesn't work when calling `context.clear()`
I am testing `"vxflw-early-access": "5.0.0-alpha.0"` and running into bizzare scaling issues. Here is the correct output at a `scale` factor of `1`.

image

image

When setting `width` and `height` like so to a `scale` factor of `0.6`:

```ts
renderer.resize(width * scale, height * scale)
renderer.getContext().scale(scale, scale)
```

The SVG output applies the correct `width`, `height` and `style` attributes. But the `viewbox` is completely messed up.

image

image

If I omit the `scale()` call:

```ts
renderer.resize(width * scale, height * scale)
//renderer.getContext().scale(scale, scale)
```

Then the `viewbox` dimensions are correct, but the image is not scaled in any way - only the bounding dimensions.

image

image

I did not experience this issue with v4. Is there a new way to handling scaling in v5?

rvilarl:
fixes #202 

@AaronDavidNewman can you suggest a test?

There are no visual differences.

### DIFF
--- a/src/svgcontext.ts
+++ b/src/svgcontext.ts
@@ -346,8 +346,6 @@ export class SVGContext extends RenderContext {
       this.svg.removeChild(this.svg.lastChild);
     }
 
-    // Replace the viewbox attribute we just removed.
-    this.scale(this.state.scaleX as number, this.state.scaleY as number);
   }
 
   // ## Rectangles:


rvilarl:
@alindsay55661 does this fix your issue?

alindsay55661:
> @alindsay55661 does this fix your issue?

Yes, I can confirm this fixes my issue.

alindsay55661 (approved):

PR #216, opened 2024-07-12
new element style infrastructure

### LINKED ISSUE #161: ElementStyle infrastructure is not consistent
Looking at the dasharray I have realized that the styles are not coded consistently.
I will try to make some PRs without visual differences, starting with Bend.

rvilarl:
This PR is a proposal to resolve #161, it also returns the grey lines to the bends.

My proposal to make the ElementStyle Infrastructure consistent is:
- Include the defaults in `metrics.ts`
- Use `this.style` together with `setStyle()` and `getStyle()` defined on `Element` for derivate classes that require only one style and avoid `render_options` related to style (see `stave.ts`)
- For classes deriving from `Element` that require more than one style, define the additional `styleXXX` (with `getStyleXXX()` and `setStyleXXX()`) in the derived class. (see `bend.ts`)

### DIFF
--- a/src/element.ts
+++ b/src/element.ts
@@ -200,13 +200,13 @@ export class Element {
    * ```
    */
   setStyle(style: ElementStyle): this {
-    this.style = {...this.style, ...style};
+    this.style = style;
     return this;
   }
 
   /** Set the element & associated children style used for rendering. */
   setGroupStyle(style: ElementStyle): this {
-    this.style = {...this.style, ...style};
+    this.style = style;
     this.children.forEach((child) => child.setGroupStyle(style));
     return this;
   }
--- a/src/element.ts
+++ b/src/element.ts
@@ -114,7 +114,7 @@ export class Element {
   protected attrs: ElementAttributes;
 
   protected rendered: boolean;
-  protected style?: ElementStyle;
+  protected style: ElementStyle = {};
   protected registry?: Registry;
 
   protected _fontInfo: Required<FontInfo>;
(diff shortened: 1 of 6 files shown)


rvilarl:
This approach allows to use all the styles. See example below which I have done just for fun (not included in test catalog)

image

src/element.ts
```diff
@@ -198,29 +199,29 @@ export class Element {
    * element.drawWithStyle();
    * ```
    */
-  setStyle(style: ElementStyle | undefined): this {
-    this.style = style;
+  setStyle(style: ElementStyle): this {
+    this.style = {...this.style, ...style};
```
mscuthbert, line 203: This is a major and breaking change that I'm not sure has been adequately discussed as far as the implications and definitely not adequately documented. It's probably okay but how would someone remove a key from this.style using the interface now? This seems to me to be a new method like "updateStyle" or something. I would not expect a method called "setStyle" to retain attributes from the previous style.

src/element.ts
```diff
@@ -198,29 +199,29 @@ export class Element {
...
     return this;
   }
 
   /** Set the element & associated children style used for rendering. */
   setGroupStyle(style: ElementStyle): this {
-    this.style = style;
+    this.style = {...this.style, ...style};
```
mscuthbert, line 209: Same as above

mscuthbert (commented):
The changes to bends and font metrics caching etc I'm +1 on. 

The changes to the meaning of setStyle and setGroupStyle I'm strongly -1 on at this stage of the v5 beta. It's a substantial behavior change at a late stage without enough discussion. 

Thanks!

src/element.ts
rvilarl: Agree, reverted. Same visual results.

src/element.ts
rvilarl: Agree, reverted. Same visual results.

rvilarl:
@mscuthbert let me know if it is fine to merge now.

mscuthbert:
> @mscuthbert let me know if it is fine to merge now.

Yes. Merged.

PR #217, opened 2024-07-12
fix curve api

### LINKED ISSUE #205: Improve Curve API
- https://github.com/0xfe/vexflow/issues/1607 (I think this is a big ask)

_Originally posted by @jaredjj3 in https://github.com/vexflow/vexflow/issues/176#issuecomment-1978655914_

rvilarl:
fixes #205 bullet 1 & 2 I do not know how to proceed with 3.

New test for opening direction

image

### DIFF
--- a/tests/curve_tests.ts
+++ b/tests/curve_tests.ts
@@ -15,7 +15,8 @@ const CurveTests = {
     QUnit.module('Curve');
     const run = VexFlowTests.runTests;
     run('Simple Curve', simple);
-    run('Opening Direction', openingDir);
+    run('Opening Direction - up', openingDirUp);
+    run('Opening Direction - down', openingDirDown);
     run('Rounded Curve', rounded);
     run('Thick Thin Curves', thickThin);
     run('Top Curve', top);
@@ -84,7 +85,7 @@ const simple = createTest(
   }
 );
 
-const openingDir = createTest(
+const openingDirUp = createTest(
   ['c4/8, f5, d5, g5', { stem: 'up' }], // beamGroup1
   ['d6/8, f5, d5, g5', { stem: 'down' }], // beamGroup2
   (f, notes) => {
@@ -92,10 +93,6 @@ const openingDir = createTest(
       from: notes[0],
       to: notes[3],
       options: {
-        cps: [
-          { x: 0, y: 10 },
-          { x: 0, y: 50 },
-        ],
         openingDirection: 'up',
       },
     });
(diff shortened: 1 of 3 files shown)


rvilarl:
@jaredjj3 could you also have a look?

tests/curve_tests.ts
```diff
@@ -83,6 +84,36 @@ const simple = createTest(
...
+  (f, notes) => {
+    f.Curve({
+      from: notes[0],
+      to: notes[3],
+      options: {
+        cps: [
+          { x: 0, y: 10 },
+          { x: 0, y: 50 },
+        ],
+        openingDirection: 'up',
```
jaredjj3, line 99: In the test image, it looks like it curves up then down. Does it look this way without those `cps`?

It seems like in order to honor `openingDirection`, vexflow will have to control `cps`. What have you found when experimenting with it?

rvilarl: Without the cps on the first curve

image

rvilarl: I have not experimented much :)

rvilarl:
@jaredjj3 is this PR fine?

jaredjj3:
While the API is what I asked for, it doesn't seem to guarantee the desired effect because of the issue with `cps`. Is there a way you can guarantee that the combination of `(openingDirection, invert)` options are honored?

rvilarl:
> While the API is what I asked for, it doesn't seem to guarantee the desired effect because of the issue with `cps`. 

@jaredjj3 could you provide a drawing of the wished result with up and down?
> Is there a way you can guarantee that the combination of `(openingDirection, invert)` options are honored?

yes, up + invert results in down and down + invert results in up

jaredjj3:
This is what I mean. However, I think `openingDirection` would be obsolete if `vexflow` suggested `cps` that rendered in a way that avoids bounding boxes of other tickables. I speculate that people would not want the second measure as an option.

image

Do you think `cps` just has to be manually adjusted by a human? After all, the [<slur> MusicXML](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/slur/) encodes that information.

rvilarl:
> This is what I mean. However, I think `openingDirection` would be obsolete if `vexflow` suggested `cps` that rendered in a way that avoids bounding boxes of other tickables. I speculate that people would not want the second measure as an option.
> 
> image

I assumed 'up' and 'down' in the other direction, now fixed. I have produced to tests one for 'up'

image

and one for 'down'

image

> Do you think `cps` just has to be manually adjusted by a human? After all, the [ MusicXML](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/slur/) encodes that information.

I think that the fact that  musicxml is storing it, indicates manual adjustment done in the edition tool.

@jaredjj3 is that now looking good?

jaredjj3:
@rvilarl, yes! This is exactly what I meant. This should let me control slur directions in vexml. LGTM

PR #219, opened 2024-07-13
Issue218

### LINKED ISSUE #218: puppeteer -> Error: Page crashed!
I am getting random `Error: Page crashed!` while generating images using `npm run test:reference`

I have found the following page https://github.com/puppeteer/puppeteer/issues/1321

adding `, args: ['--disable-gpu', '--no-sandbox', '--lang=en-US', '--disable-setuid-sandbox', '--disable-dev-shm-usage']` to `puppeteer.launch` seems to resolve the problem.

rvilarl:
fixes #218 

npm update
revert to jsdom as default.

No visual differences.


rvilarl:
@ronyeh I have to merge this one to be able to check differences. Please look into it. If there is something wrong we can revert.

ronyeh:
No problem! If it fixes a crash, I trust your judgment.

Can you provide a little more detail on what was going on, for future
reference? We should treat our issues and PRs as form of documentation of
our thought process.

tools/generate_images.js
```diff
@@ -81,7 +81,7 @@ const parseArgs = () => {
   });
 
   backends = backends ?? {
-    pptr: true,
+    jsdom: true,
```
ronyeh, line 84: Maybe instead of deleting the pptr line, we can comment it out? These scripts are not well documented so it might be an easy way to show which flags are available.

ronyeh (commented):
@rvilarl  added a small comment about documentation (or commenting out the pptr flag instead of removing it).

tools/generate_images.js
rvilarl: The options are in the code a few lines above

https://github.com/vexflow/vexflow/blob/7f871dcc05856de626ec91bf8d29d590eb6dda28/tools/generate_images.js#L53-L58

also in the help

https://github.com/vexflow/vexflow/blob/7f871dcc05856de626ec91bf8d29d590eb6dda28/tools/generate_images.js#L13-L21

rvilarl:
> No problem! If it fixes a crash, I trust your judgment. Can you provide a little more detail on what was going on, for future reference? We should treat our issues and PRs as form of documentation of our thought process.

comments added to #218

tools/generate_images.js
rvilarl: @ronyeh is the above fine or do you want me to add a comment?

PR #220, opened 2024-07-13
Issue194 - ctx.applyStyle()/removeStyle() should always push a state

### LINKED ISSUE #194: ctx.applyStyle()/removeStyle() should always push a state
From #189 discussion (edited for tense)

> Make applyStyle() and removeStyle() go ahead and do a ctx.save()/ctx.restore() if there is a context defined whether or not style is defined.  It seems like the small overhead of keeping an extra state-push/pop is worth it for avoiding the subtle bugs that occur when state isn't fully restored.

Endorsed by @AaronDavidNewman 

Entails moving the check for defined `this.style` to after `setState()`/`restoreState()` and removing `setState()`/`restoreState()` from methods that also call `applyStyle()`/`removeStyle()` in their code.

rvilarl:
Fixes #194 

This is the proposal that I described in #194

No visual differences.

### DIFF
--- a/src/bend.ts
+++ b/src/bend.ts
@@ -171,6 +171,7 @@ export class Bend extends Modifier {
     const ctx = this.checkContext();
     const note = this.checkAttachedNote();
     this.setRendered();
+    ctx.save();
     this.applyStyle();
 
     const start = note.getModifierStartXY(Modifier.Position.RIGHT, this.index);
@@ -189,21 +190,23 @@ export class Bend extends Modifier {
       const cpX = x + width;
       const cpY = y;
 
+      ctx.save();
       this.applyStyle(ctx, this.styleLine);
       ctx.beginPath();
       ctx.moveTo(x, y);
       ctx.quadraticCurveTo(cpX, cpY, x + width, height);
       ctx.stroke();
-      this.restoreStyle(ctx, this.styleLine);
+      ctx.restore();
     };
 
     const renderRelease = (x: number, y: number, width: number, height: number) => {
+      ctx.save();
       this.applyStyle(ctx, this.styleLine);
       ctx.beginPath();
       ctx.moveTo(x, height);
       ctx.quadraticCurveTo(x + width, height, x + width, y);
       ctx.stroke();
-      this.restoreStyle(ctx, this.styleLine);
+      ctx.restore();
     };
 
     const renderArrowHead = (x: number, y: number, direction: number) => {
@@ -280,6 +283,6 @@ export class Bend extends Modifier {
     } else if (lastBend.type === Bend.DOWN) {
       renderArrowHead(lastBend.x + lastDrawnWidth, start.y, -1);
     }
-    this.restoreStyle();
+    ctx.restore();
   }
 }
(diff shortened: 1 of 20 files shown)


rvilarl:
This exercise has made me realise that most of elements apply style already on ´draw()´. What is the need of `drawWithStyle()`then? Should it be deleted?

rvilarl:
Once this PR is in I will continue with the standardization of style as started in #216

rvilarl:
I will simplify this PR to facilitate the review. I will:
- replace restoreStyle with save
- replace applyStyle with save + applyStyle
- delete restoreStyle
- delete the call to save in ApplyStyle

rvilarl:
Done, it should be easier to review.

src/bend.ts
```diff
@@ -280,6 +283,6 @@ export class Bend extends Modifier {
     } else if (lastBend.type === Bend.DOWN) {
       renderArrowHead(lastBend.x + lastDrawnWidth, start.y, -1);
     }
-    this.restoreStyle();
+    ctx.restore();
```
ronyeh, line 286: Is this ctx.restore() paired with the ctx.save() from line 174? It's a big method so I wanted to make sure the save/restore calls are all correctly matched.

src/element.ts
```diff
@@ -218,10 +218,7 @@ export class Element {
 
   /** Apply the element style to `context`. */
   applyStyle(context: RenderContext | undefined = this.context, style: ElementStyle = this.getStyle()): this {
-    if (Object.keys(style).length == 0) return this;
     if (!context) return this;
-
-    context.save();
```
ronyeh, line 224: I'm OK with this new approach. The only risk is if someone manually called applyStyle(...) in their v4 compatible code, they'll expect that context.save() has been called. But with this new approach, it won't be called.

But we're removing restoreStyle() below, so anyone who manually calls applyStyle() / restoreStyle() will notice very quickly that their code is broken since it won't compile.

ronyeh (approved):
Looks good to me. Please see my questions / comments above.

src/bend.ts
mscuthbert: I don't think it's such a big deal, but if we cared about making sure things were correctly matched, we could make ctx.save() return an object like:

```
const savedContext = ctx.save();
...
savedContext.restore();
```

where ctx.save returns:
```
 {
  restore:  () => { 
    if (this._ctx) { 
      this._ctx.restore(); 
      this._ctx = null; 
    }
  },
  _ctx: this,
}
```

Then we don't have to worry about a context being restored twice, or from a non-existent save, and failing to restore context will give a typescript warning that `savedContext` was defined but never used.

rvilarl: @ronyeh thanks a lot for your review! yes, this ctx.restore() is paired with the ctx.save() from line 174
@mscuthbert the canvas context does not return a value, see https://www.w3schools.com/jsref/canvas_save.asp. I believe that the svg context should not be have differently.

My proposal in #223 reduces a lot the number of calls to ´ctx.save()´ I will raise a PR.

src/element.ts
rvilarl: Agree

PR #225, opened 2024-07-25
Refactor calls to applyStyle

### LINKED ISSUE #223: ApplyStyle in derived classes from Element
This exercise has made me realise that some elements apply style already on `draw()`. What is the need of `drawWithStyle()`then? Should it be deleted?
@mscuthbert @AaronDavidNewman @ronyeh what do you think should it drawWithStyle be deleted or should we refactor the style out of draw?

_Originally posted by @rvilarl in https://github.com/vexflow/vexflow/issues/220#issuecomment-2226974080_

rvilarl:
Fixes #223
This PR reduces the number of calls to context `save` and `restore`. It calls consistently to `drawWithStyle()` rather than `draw()`.

With this approach there is no longer missing calls to applyStyle on some elements (ie. Voice) and double calls are also avoided.

One visual difference see #226

### DIFF
--- a/src/element.ts
+++ b/src/element.ts
@@ -170,7 +170,7 @@ export class Element {
    * Note that StaveNote calls setGroupStyle() when setStyle() is called.
    */
   addChildElement(child: Element): this {
-    if (child.parent) throw new RuntimeError('Element', 'Draw not defined');
+    if (child.parent) throw new RuntimeError('Element', 'Parent already defined');
     child.parent = this;
     this.children.push(child);
     return this;
--- a/src/element.ts
+++ b/src/element.ts
@@ -81,6 +81,7 @@ export class Element {
 
   // Element objects keep a list of children that they are responsible for.
   // Children inherit the style from their parents (see: setGroupStyle(s)).
+  protected parent?: Element;
   protected children: Element[] = [];
   protected static ID: number = 1000;
   protected static newID(): string {
@@ -169,6 +170,8 @@ export class Element {
    * Note that StaveNote calls setGroupStyle() when setStyle() is called.
    */
   addChildElement(child: Element): this {
+    if (child.parent) throw new RuntimeError('Element', 'Draw not defined');
+    child.parent = this;
     this.children.push(child);
     return this;
   }
@@ -592,7 +595,6 @@ export class Element {
 
   /** Render the element text. */
   renderText(ctx: RenderContext, xPos: number, yPos: number): void {
-    ctx.save();
     ctx.setFont(this._fontInfo);
     ctx.fillText(this._text, xPos + this.x + this.xShift, yPos + this.y + this.yShift);
     this.children.forEach((child) => {
@@ -600,7 +602,6 @@ export class Element {
       ctx.setFont(child.fontInfo);
       ctx.fillText(child.text, xPos + child.x + child.xShift, yPos + child.y + child.yShift);
     });
-    ctx.restore();
   }
 
   /** Measure the text using the textFont. */
--- a/src/element.ts
+++ b/src/element.ts
@@ -161,7 +161,7 @@ export class Element {
   }
 
   /**
-   * Adds a child Element to the Element, which lets it inherit the
+   * Adds a child to the Element, which lets it inherit the
    * same style as the parent when setGroupStyle() is called.
    *
    * Examples of children are noteheads and stems.  Modifiers such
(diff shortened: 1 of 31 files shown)


rvilarl:
I have realised that this is a lot of changes. I will make smaller PRs which should be easier/quicker to review and will reduce the number of changed files in this one.

rvilarl:
@ronyeh it might make sense to finish that. But it is not a must for 5.0

src/element.ts
```diff
@@ -169,6 +170,8 @@ export class Element {
    * Note that StaveNote calls setGroupStyle() when setStyle() is called.
    */
   addChildElement(child: Element): this {
+    if (child.parent) throw new RuntimeError('Element', 'Draw not defined');
```
ronyeh, line 173: I don't understand this error message. How is it related to draw when you're checking that child already has a parent?

ronyeh (commented):
After this PR, are there any more calls to draw? Or does it always go through drawWithStyle now?

src/element.ts
rvilarl: Good catch. The comment is wrong.

rvilarl: Error message changed

rvilarl:
> After this PR, are there any more calls to draw? Or does it always go through drawWithStyle now?

I believe that everything goes through drawWithStyle.

src/element.ts
```diff
@@ -169,6 +170,8 @@ export class Element {
    * Note that StaveNote calls setGroupStyle() when setStyle() is called.
    */
   addChildElement(child: Element): this {
```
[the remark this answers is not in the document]
ronyeh: It was introduced with this commit:
https://github.com/vexflow/vexflow/commit/da383af131afc9995b63effdde67831d5e5999d5

There is no other addChild in our VexFlow repository. 

Could it be simplified by renaming it to addChild instead of addChildElement?

src/staveline.ts
```diff
@@ -261,15 +261,14 @@ export class StaveLine extends Element {
   }
 
   // Renders the `StaveLine` on the context
-  draw(): this {
+  draw(): void {
```
ronyeh, line 264: There is also a `draw(): this` in `stavesection.ts`.

Should that one also be changed to void? In the repo, draw() almost always returns void.

src/stavenote.ts
```diff
@@ -1113,24 +1109,18 @@ export class StaveNote extends StemmableNote {
...
 
       // Draw the Flag
-      ctx.save();
-      this.applyStyle(ctx, this.flagStyle);
-      this.flag.renderText(ctx, flagX, flagY);
-      ctx.restore();
+      this.flag.draw = (): void => {
+        this.flag.renderText(ctx, flagX, flagY);
+      };
+      this.flag.setContext(ctx).drawWithStyle();
```
ronyeh, line 1115: Interesting that we define this draw function on the fly.

Should the flag be its own class?

src/tabnote.ts
```diff
@@ -351,10 +351,10 @@ export class TabNote extends StemmableNote {
...
 
       // Draw the Flag
-      context.save();
-      this.applyStyle(context, this.flagStyle);
-      this.flag.renderText(context, flagX, flagY);
-      context.restore();
+      this.flag.draw = (): void => {
+        this.flag.renderText(context, flagX, flagY);
+      };
+      this.flag.setContext(context).drawWithStyle();
```
ronyeh, line 357: Same here. Should the flag be its own class? That way we don't have to define the draw function like this.

ronyeh (changes requested):
See comments. Let me know what you think. I can merge after these questions are answered or addressed. Thanks!

src/element.ts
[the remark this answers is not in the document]
rvilarl: done

src/stavenote.ts
rvilarl: done

src/staveline.ts
rvilarl: done

src/stavenote.ts
rvilarl: done

src/tabnote.ts
rvilarl: done

src/flag.ts
```diff
@@ -0,0 +1,35 @@
...
+function L(...args: any[]) {
+  if (Flag.DEBUG) log('VexFlow.Flag', args);
+}
+
+/**
+ * `Flags` are typically not manipulated
+ * directly, but used internally in `StaveNote`.
+ */
+export class Flag extends Element {
+  /** To enable logging for this class. Set `VexFlow.NoteHead.DEBUG` to `true`. */
```
ronyeh, line 18: I can merge this but please fix the comment above at some point.

ronyeh (approved):
Just one comment to fix.

PR #227, opened 2024-07-25
refactor applyStyle on Modifiers (part of #225)

rvilarl:
This is part of #225:
- avoids duplicated calls to ctx.save on Modifiers

No visual differences.

### DIFF
--- a/src/annotation.ts
+++ b/src/annotation.ts
@@ -224,8 +224,6 @@ export class Annotation extends Modifier {
     const start = note.getModifierStartXY(ModifierPosition.ABOVE, this.index);
 
     this.setRendered();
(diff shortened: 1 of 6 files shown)


ronyeh (approved):

ronyeh:
Let me know if I'm understanding this correctly.

In VexFlow 4.x, draw() will apply styles to some modifiers / elements, but not all of them.

With your new PRs, draw() will no longer apply any styles to any modifier / element.

To apply a style, you need to call drawWithStyle().

Is that right?

rvilarl:
Yes this is right. With that we will also avoid double calls and sometimes no call at all to applyStyle. see #223

PR #228, opened 2024-07-25
readme update for release

rvilarl:
just a readme update with focus on releasing

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ The quickest way to add VexFlow to a web page is via a `<script>` tag.
 <div id="output"></div>
 
 <!-- Load library -->
-<script src="https://cdn.jsdelivr.net/npm/vxflw-early-access@alpha/build/cjs/vexflow.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/vexflow@5.0.0-alpha.4/build/cjs/vexflow.js"></script>
 
 <script>
   document.fonts.ready.then(() => {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -73,13 +73,13 @@ system
...
 
-[See it running here.](https://vexflow.github.io/vexflow-examples/static/step0)
+[See it running here.](https://vexflow.github.io/vexflow-examples/getting-started)
 
 [Learn more about EasyScore here.](https://github.com/0xfe/vexflow/wiki/Using-EasyScore)
 
 ## Native API
 
-If you need more control, you can use the low-level VexFlow API. Below, we render a stave using SVG. [See a running example of the low-level API here.](https://jsfiddle.net/5zgf03un/)
+If you need more control, you can use the low-level VexFlow API. Below, we render a stave using SVG. [See a running example of the low-level API here.](https://vexflow.github.io/vexflow-examples/getting-started)
```
ronyeh, line 82: Line 82 and Line 76 refer to the same URL. Should they refer to a different URL (or at least a different `#section` since they are talking about different examples?

ronyeh: There is also a typo on the vexflow-examples page. It says "It is also possible to use face fonts"

rvilarl: But this is still possible, is not it? Perhaps we should reduce the README.md just to go directly into the vexflow-example sections and avoid the double maintenance of examples.

sschmidTU: @rvilarl What is a face font? :)

rvilarl: font-face

PR #230, opened 2024-08-04
StaveSection with configurable padding

### LINKED ISSUE #63: Stave section rect needs more internal white space padding all around.
This can be a separate PR, but I think the rect needs more internal padding all around. Right now it crowds the letter B in all 4 directions. :-)

_Originally posted by @ronyeh in https://github.com/vexflow/vexflow/issues/50#issuecomment-1684881222_

rvilarl:
fixes #63 

Not required for 5.0

Visual differences

current (section B with extra padding)
Stave Multiple_Staves_Volta_Test Bravura jsdom_current

reference
Stave Multiple_Staves_Volta_Test Bravura jsdom_reference

### DIFF
--- a/tests/stave_tests.ts
+++ b/tests/stave_tests.ts
@@ -152,7 +152,7 @@ function drawMultipleMeasures(options: TestOptions, contextBuilder: ContextBuild
   const staveBar1 = new Stave(10, 50, 200);
   staveBar1.setBegBarType(BarlineType.REPEAT_BEGIN);
   staveBar1.setEndBarType(BarlineType.DOUBLE);
-  staveBar1.addModifier(new StaveSection('A').setFontSize(options.params?.fontSize).setDrawRect(false));
+  staveBar1.setSection('A', 0, 0, options.params?.fontSize, false);
   staveBar1.addClef('treble').setContext(ctx).draw();
   const notesBar1 = [
     new StaveNote({ keys: ['c/4'], duration: 'q' }),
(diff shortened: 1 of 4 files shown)


tests/stave_tests.ts
```diff
@@ -640,7 +641,7 @@ function drawVoltaModifier(options: TestOptions, contextBuilder: ContextBuilder)
   // mm5.addModifier(new Repetition(Repetition.type.DS, mm4.getX() + mm4.getWidth(), 50), StaveModifierPosition.RIGHT);
   mm6.setRepetitionType(Repetition.type.DS);
   mm6.setMeasure(6);
-  mm6.setSection('E', 0);
+  mm6.addModifier(new StaveSection('E'));
```
ronyeh, line 644: I noticed all the calls to .setSection() have been replaced with addModifier(new StaveSection(...)).

Does the older API still work? setSection has been around since VexFlow 3.0.9 so hopefully the older approach still works (with good defaults).

Do we have a wiki document to document such changes? That way, people coming from VexFlow 4.x or 3.x can know that .addModifier(new StaveSection(...)) is the newer preferred approach. (Although I think the previous API seems simpler, and should be kept around).

ronyeh: Here is the API in VexFlow 3.0.9:

https://github.com/0xfe/vexflow/blob/3.0.9/tests/stave_tests.js

```js
var staveBar1 = new VF.Stave(10, 50, 200);
staveBar1.setBegBarType(VF.Barline.type.REPEAT_BEGIN);
staveBar1.setEndBarType(VF.Barline.type.DOUBLE);
staveBar1.setSection('A', 0);
staveBar1.addClef('treble').setContext(ctx).draw();
```

rvilarl: Calling setSection does not provide any pointer the StaveSection and this why I decided to use the standard addModifier. setSection is still available but does not allow to change the padding.

ronyeh:
If setSection is still available, is it tested anywhere? It looks like we
replaced them all with addModifier. I just want to make sure setSection
still works properly. If we don’t test it, it might silently fail in some
future PR.

tests/stave_tests.ts
rvilarl: setSection now also being tested.

rvilarl:
ready to merge

ronyeh (approved):

ronyeh:
Thanks!

PR #233, opened 2024-08-06
tuplet text position configurable

### LINKED ISSUE #62: Tuplet vertical alignment of number with bracket
See: https://github.com/vexflow/vexflow/pull/55

My only concern is that the brackets might be a tiny tiny bit too low?

The reference images seem more vertically centered to me, but in the current images, the number 3 seems like it could be higher up.

I think the pointy part of the number 3 should be vertically aligned with the horizontal line of the bracket. Or maybe the pointy part of the 3 could even be a tiny bit higher.

Maybe we should pick a good default, but make it customizable (at some point)?

_Originally posted by @ronyeh in https://github.com/vexflow/vexflow/issues/55#issuecomment-1684356354_

rvilarl:
fixes #62

Not required for 5.0

Visual differences
current
Auto_Beaming Duration_Based_Secondary_Beam_Breaks_2 Bravura jsdom_current
reference
Auto_Beaming Duration_Based_Secondary_Beam_Breaks_2 Bravura jsdom_reference
current
Auto_Beaming More_Simple_Tuplet_Auto_Beaming Bravura jsdom_current
reference
Auto_Beaming More_Simple_Tuplet_Auto_Beaming Bravura jsdom_reference

### DIFF
--- a/src/tuplet.ts
+++ b/src/tuplet.ts
@@ -47,6 +47,7 @@
 import { Element } from './element';
 import { Formatter } from './formatter';
 import { Glyphs } from './glyphs';
+import { Metrics } from './metrics';
 import { Note } from './note';
 import { Stem } from './stem';
 import { StemmableNote } from './stemmablenote';
@@ -61,13 +62,7 @@ export interface TupletOptions {
   numNotes?: number;
   ratioed?: boolean;
   yOffset?: number;
-}
-
-export interface TupletMetrics {
-  noteHeadOffset: number;
-  stemOffset: number;
-  bottomLine: number;
-  topModifierOffset: number;
+  textYOffset?: number;
 }
 
 export const enum TupletLocation {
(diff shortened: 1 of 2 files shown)


src/tuplet.ts
```diff
@@ -150,15 +140,15 @@ export class Tuplet extends Element {
    * Set whether or not the bracket is drawn.
    */
   setBracketed(bracketed: boolean): this {
-    this.bracketed = !!bracketed;
+    this.options.bracketed = !!bracketed;
```
ronyeh, line 143: Do we still need to use the not not here to force the param to be boolean? I think it was a vestige of classic JS (see VexFlow 3.0.9 at https://github.com/0xfe/vexflow/blob/00ec15c67ff333ea49f4d3defbd9e22374c03684/src/tuplet.js#L123).

src/tuplet.ts
```diff
@@ -150,15 +140,15 @@ export class Tuplet extends Element {
...
     return this;
   }
 
   /**
    * Set whether or not the ratio is shown.
    */
   setRatioed(ratioed: boolean): this {
-    this.ratioed = !!ratioed;
+    this.options.ratioed = !!ratioed;
```
ronyeh, line 151: Same here. I'm okay with removing the !! hack to force the parameter to be a boolean.

ronyeh (changes requested):

PR #235, opened 2024-08-07
grunt eslint

ronyeh:
When I ran `grunt eslint` on the most recent version, I get these changes.

Any idea why eslint is removing these directives?

```
// eslint-disable-next-line
```

### DIFF
--- a/tests/accidental_tests.ts
+++ b/tests/accidental_tests.ts
@@ -338,13 +338,12 @@ function genAccidentals(): string[] {
   }
   // Spartan Sagittal multi-shaft accidentals
   for (let u = 0xe310; u <= 0xe335; u++) {
-    switch (
-      u // exclude unused smufls
-    ) {
+    switch (u) {
       case 0xe31a:
       case 0xe31b:
       case 0xe31e:
       case 0xe31f:
+        // Exclude unused SMuFL glyphs.
         break;
       default:
         accs.push(String.fromCodePoint(u));
--- a/tests/accidental_tests.ts
+++ b/tests/accidental_tests.ts
@@ -338,7 +338,9 @@ function genAccidentals(): string[] {
   }
   // Spartan Sagittal multi-shaft accidentals
   for (let u = 0xe310; u <= 0xe335; u++) {
-    switch (u) {  // exclude unused smufls
+    switch (
+      u // exclude unused smufls
+    ) {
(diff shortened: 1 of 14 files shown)


tests/accidental_tests.ts
```diff
@@ -338,7 +338,9 @@ function genAccidentals(): string[] {
   }
   // Spartan Sagittal multi-shaft accidentals
   for (let u = 0xe310; u <= 0xe335; u++) {
-    switch (u) {  // exclude unused smufls
+    switch (
+      u // exclude unused smufls
```
ronyeh, line 342: I see this weird comment style is caused by eslint.

tests/accidental_tests.ts
```diff
@@ -410,10 +412,10 @@ function cautionary(options: TestOptions): void {
 
   const accids = Object.values(accidentals).filter((accid) => accid !== '{' && accid !== '}');
   for (i = 0; i < staveCount; ++i) {
-    const stave = f.Stave({ x: 0, y: 10 + staveHeight / scale * i, width: staveWidth / scale });
+    const stave = f.Stave({ x: 0, y: 10 + (staveHeight / scale) * i, width: staveWidth / scale });
     const score = f.EasyScore();
     const rowMap = [];
-    for (j = 0; j < notesPerStave && (j + i * notesPerStave) < accids.length; ++j) {
+    for (j = 0; j < notesPerStave && j + i * notesPerStave < accids.length; ++j) {
```
ronyeh, line 418: Also, it looks like the weird parentheses changes are caused by eslint.

rvilarl: The first one tries to avoid confusion with a / (b*c)
The second is that eslint considers parenthesis for the whole expression unnecessary.

rvilarl:
> When I ran `grunt eslint` on the most recent version, I get these changes.
> 
> Any idea why eslint is removing these directives?
> 
> ```
> // eslint-disable-next-line
> ```

I was also surprised but it seems that they are not required, the modified code compiles without complainig.

ronyeh:
In the interest of moving forward, I'm going to merge this PR which is caused by running `grunt eslint`. If we ever figure out what caused eslint to remove all our overrides, we can add those comments below. (Maybe our eslint rules are more relaxed now? Or did eslint change their default rules?)

PR #236, opened 2024-08-08
embed svg font

rvilarl:
This PR fixes the svg node demo and provides a python script to embed the required subset of the font.

To try it run:
```sh
node svg.ts > test.svg
pyhton3 svgftembed.py test.svg bravura
```

I think that it would make sense to make another script to transform font glyphs to paths using opentypejs

### DIFF
--- a/demos/node/svg.js
+++ b/demos/node/svg.js
@@ -15,11 +15,8 @@
 // Additionally, we could explore SVG <use> so we can define one template path and reuse it multiple times in one score.
 // See: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use
 
-// const Vex = require('vexflow');
-const Vex = require('vxflw-early-access');
+const VF = require('vexflow');
 const { JSDOM } = require('jsdom');
-
-const VF = VexFlow;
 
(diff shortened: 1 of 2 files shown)


rvilarl:
@jsmcx how do you create the svgs. What steps do you folow? I ask in order to automatize the conversion.

jsmcx:
> @jsmcx how do you create the svgs. What steps do you folow? I ask in order to automatize the conversion.

This was just a proof of concept that was written by hand. I'm sure there are tools to do it properly. 

The steps were:
- Decode the base64 subset font in your examples, and extract the glyph for the treble clef
- Wrap the paths in a `<symbol>` tag within SVG `<defs>`
- Apply a transformation to flip the image horizontally and adjust the baseline (fonts use a different coordinate system to SVG)
- Instantiate with `<use>` tags in the body of the SVG

ronyeh:
This is useful example code, but I think we need to figure out how to solve it in a way that doesn't require changes to the svg.js demo.

These mini demos illustrate the external API of vexflow, that users have become accustomed to. Requiring users (i.e. developers) to write a bunch of extra code to handle the SVG adds extra burden on them. They are already busy dealing with their own apps. VexFlow should do the heavy lifting and make sure SVG output works without any surprises.

PR #237, opened 2024-08-11
keysignature refactor boundingbox calculations

### LINKED ISSUE #121: KeySignature as Element
Refactor format & width calculation. Accidentals should be child elements.

### LINKED ISSUE #131: Key signatures too close
see https://github.com/stringsync/vexml/pull/152#issuecomment-1763785453

rvilarl:
fixes #121
fixes #131

not required for 5.0
the refactoring has revealed that the bounding box calculation was wrong.

Visual differences:
- first test now includes boundig boxes.
- keysignatures leave a bit more space for other elements in the stave

current
KeySignature Major_Key_Test Bravura jsdom_current
reference
KeySignature Major_Key_Test Bravura jsdom_reference

### DIFF
--- a/src/keysignature.ts
+++ b/src/keysignature.ts
@@ -8,6 +8,7 @@
 
 import { BoundingBox } from './boundingbox';
 import { Element } from './element';
+import { Glyphs } from './glyphs';
 import { Stave } from './stave';
 import { StaveModifier, StaveModifierPosition } from './stavemodifier';
 import { Tables } from './tables';
@@ -49,11 +50,16 @@ export class KeySignature extends StaveModifier {
     glyph.setText(code);
 
     // Determine spacing between current accidental and the next accidental
-    const extraWidth = 1;
+    let extraWidth = 1;
     // Place the glyph on the stave
     glyph.setYShift(stave.getYForLine(acc.line));
     if (this.children.length > 0) {
       const prevGlyph = this.children[this.children.length - 1];
+      if (
+        (prevGlyph.getText() === Glyphs.accidentalNatural || glyph.getText() === Glyphs.accidentalNatural) &&
+        Math.abs(glyph.getYShift() - prevGlyph.getYShift()) < 10
+      )
+        extraWidth = 2;
       glyph.setXShift(prevGlyph.getXShift() + prevGlyph.getWidth() + extraWidth);
     }
     this.children.push(glyph);
--- a/src/keysignature.ts
+++ b/src/keysignature.ts
@@ -55,10 +55,10 @@ export class KeySignature extends StaveModifier {
     glyph.setYShift(stave.getYForLine(acc.line));
     if (this.children.length > 0) {
       const prevGlyph = this.children[this.children.length - 1];
-      if (
-        (prevGlyph.getText() === Glyphs.accidentalNatural || glyph.getText() === Glyphs.accidentalNatural) &&
-        Math.abs(glyph.getYShift() - prevGlyph.getYShift()) < 10
-      ) {
+      const isNatural = (el: Element) => el.getText() === Glyphs.accidentalNatural;
+      const yShiftDiff = (el1: Element, el2: Element) => Math.abs(el2.getYShift() - el1.getYShift());
+
+      if ((isNatural(prevGlyph) || isNatural(glyph)) && yShiftDiff(prevGlyph, glyph) < 10) {
         extraWidth = 2;
       }
       glyph.setXShift(prevGlyph.getXShift() + prevGlyph.getWidth() + extraWidth);
--- a/src/keysignature.ts
+++ b/src/keysignature.ts
@@ -58,8 +58,9 @@ export class KeySignature extends StaveModifier {
       if (
         (prevGlyph.getText() === Glyphs.accidentalNatural || glyph.getText() === Glyphs.accidentalNatural) &&
         Math.abs(glyph.getYShift() - prevGlyph.getYShift()) < 10
-      )
+      ) {
         extraWidth = 2;
+      }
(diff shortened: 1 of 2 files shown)


rvilarl:
The last commit also fixes #131

current
KeySignature Cancelled_key__for_each_clef__test Bravura jsdom_current
reference
KeySignature Cancelled_key__for_each_clef__test Bravura jsdom_reference
current
KeySignature Key_Signature_Change_test Bravura jsdom_current
reference
KeySignature Key_Signature_Change_test Bravura jsdom_reference

src/keysignature.ts
```diff
@@ -32,27 +33,36 @@ export class KeySignature extends StaveModifier {
...
     glyph.setYShift(stave.getYForLine(acc.line));
-    if (this.glyphs.length > 0) {
-      const prevGlyph = this.glyphs[this.glyphs.length - 1];
+    if (this.children.length > 0) {
+      const prevGlyph = this.children[this.children.length - 1];
+      if (
+        (prevGlyph.getText() === Glyphs.accidentalNatural || glyph.getText() === Glyphs.accidentalNatural) &&
+        Math.abs(glyph.getYShift() - prevGlyph.getYShift()) < 10
+      )
+        extraWidth = 2;
```
ronyeh, line 62: The indentation looks weird here.

Is it because you didn't add braces { }?

We should always use braces because it improves readability and reduces bugs. Maybe we should add that rule to the eslint, but for now I'd recommend using our best judgment and generally avoiding a single line if statement without braces unless there's a great reason for it.

src/keysignature.ts
```diff
@@ -102,16 +115,37 @@ export class KeySignature extends StaveModifier {
...
+    if (!this.formatted) this.format();
+
+    return super.getBoundingBox();
+  }
+
+  // Calculate the width and height for the entire signature
+  protected calculateDimensions(): void {
+    let boundingBox: BoundingBox;
+    if (this.children.length > 0) boundingBox = this.children[0].getBoundingBox();
+    else boundingBox = new BoundingBox(this.x + this.xShift, this.y + this.yShift, 0, 0);
```
ronyeh, line 137: Same here.

I know it looks nice and compact, but I'm not a fan of this style.

Using braces on if/else blocks allows us to add new lines of code in the future without accidentally causing bugs.

I also personally think it is easier to read with braces in general.

src/keysignature.ts
```diff
@@ -158,18 +192,21 @@ export class KeySignature extends StaveModifier {
...
     if (!this.formatted) this.format();
 
-    return this.glyphs.length === 0 || (!this.paddingForced && index < 2) ? 0 : this.padding;
+    return this.children.length === 0 || (!this.paddingForced && index < 2) ? 0 : this.padding;
   }
 
+  // Get the width of the modifier
   getWidth(): number {
-    if (!this.formatted) this.calculateWidth();
+    if (!this.formatted) this.format();
```
ronyeh, line 204: This one here is probably OK because it's trivial. No need for braces. But if we ever add an eslint rule to enforce braces, then I'm fine with braces showing up here too.

src/keysignature.ts
rvilarl: done, the indentation is the one from eslint

rvilarl: done, the indentation is the one from eslint

src/keysignature.ts
rvilarl: done

src/keysignature.ts
rvilarl: I left it like it is as there are quite a few instances of
```
if (!this.formatted) this.format();
```
in the file. If you want, I can change all of them.

ronyeh:
Can you resolve the new conflicts? Thanks.

src/keysignature.ts
```diff
@@ -32,27 +33,37 @@ export class KeySignature extends StaveModifier {
...
-    if (this.glyphs.length > 0) {
-      const prevGlyph = this.glyphs[this.glyphs.length - 1];
+    if (this.children.length > 0) {
+      const prevGlyph = this.children[this.children.length - 1];
+      if (
+        (prevGlyph.getText() === Glyphs.accidentalNatural || glyph.getText() === Glyphs.accidentalNatural) &&
+        Math.abs(glyph.getYShift() - prevGlyph.getYShift()) < 10
+      ) {
+        extraWidth = 2;
+      }
```
ronyeh, line 63: The conditional is a bit difficult to read.

Can we extract them out into temporarly local variables?

Something like:
```
const prevGlyphIsNatural = prevGlyph.getText() === Glyphs.accidentalNatural;
const currGlyphIsNatural = glyph.getText() === Glyphs.accidentalNatural;
const glyphsAreTooClose = Math.abs(glyph.getYShift() - prevGlyph.getYShift()) < 10;
```

I don't know what are the best variable names here. You can choose.

src/keysignature.ts
```diff
@@ -102,16 +116,40 @@ export class KeySignature extends StaveModifier {
...
   addToStave(stave: Stave): this {
     this.paddingForced = true;
     stave.addModifier(this);
 
     return this;
   }
 
+  // Get the `BoundingBox`
+  getBoundingBox(): BoundingBox {
+    if (!this.formatted) this.format();
```
ronyeh, line 129: These one liners are completely fine in my opinion, because they're perfectly clear.

However it hurts readability when we do a one line if statement when the conditional and the statement are long and convoluted. It results in extra indentation and makes it prone to future bugs.

src/keysignature.ts
rvilarl: I have defined to local functions: `isNatural` and `yShiftDiff`

src/keysignature.ts
rvilarl: ok

PR #238, opened 2024-08-14
Grace notes accidentals' size and right position

### LINKED ISSUE #174: Gracenotes
@rvilarl, when you get a chance, take a look at the `vexml` rendering in the first comment. There are two things I notice:

1. Setting a position on a `vexflow.GraceNoteGroup` doesn't seem to affect it. For succeeding grace notes, I'm setting the position to `vexflow.ModifierPosition.RIGHT`, but it's still being rendered on the left (last note in measure 2).

2. When an accidental is set on a grace note, the scale seems a little large for it — especially when compared to osmd and MuseScore 4.

I think these are both very minor issues, so I thought you'd be interested.

_Originally posted by @jaredjj3 in https://github.com/stringsync/vexml/issues/195#issuecomment-1876182238_

rvilarl:
fixes #174 

Not required for 5.0

Accidentals on grace notes have now the same as the notes (see grace notes on note 1 & 4)
It is also allowed to place grace notes on the right (see note 5).
@jaredjj3 does this address your comments?

current
Grace_Notes Grace_Note_Basic Bravura jsdom_current
reference
Grace_Notes Grace_Note_Basic Bravura jsdom_reference

### DIFF
--- a/src/gracenotegroup.ts
+++ b/src/gracenotegroup.ts
@@ -53,7 +53,7 @@ export class GraceNoteGroup extends Modifier {
 
     const groupList = [];
     let prevNote = null;
-    let shiftL = 0;
+    let shift = 0;
 
     for (let i = 0; i < gracenoteGroups.length; ++i) {
       const gracenoteGroup = gracenoteGroups[i];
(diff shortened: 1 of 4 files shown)


jaredjj3:
Yes, it looks great, thank you!

src/gracenotegroup.ts
```diff
@@ -90,7 +94,8 @@ export class GraceNoteGroup extends Modifier {
       );
     }
 
-    state.leftShift += groupShift;
+    if (right) state.rightShift += groupShift;
+    if (left) state.leftShift += groupShift;
```
ronyeh, line 98: I think the changes in lines 78-98 could be improved.

Is the default situation `left == true` and `right == false`? We can never have both left and right have the same value, right?

Why not just have one flag `let left = true; // put the grace note group on the left side`

Then if we ever see Modifier.Position.RIGHT, we set left to false.

Then on lines 97-98, just do one or the other depending on the `left` flag.

rvilarl: Hi Ron, I have to look into that. But I guess that theoretically it is possible to have two voices each one with grace notes and one to the right and one to the left.

ronyeh: I see. If they aren't mutually exclusive then it's fine to have two separate flags. Can we add a test case that shows a note group with grace notes on both sides?

rvilarl: I will add a test case on another PR

PR #244, opened 2024-09-12
Remove references to vxflw-early-access.

ronyeh:


rvilarl:
@ronyeh  I think  that this change will use vexflow 4 until we release 5.0, do you agree?  Still the only possibility to use 5 is vexflow@beta, is not it?

ronyeh:
That’s fine by me. The demos will work once vexflow 5 is released.

ronyeh:
In fact the demos should work now. Even if they are pointing to vexflow 4.

I forget, if we run grunt to build a local copy of vexflow 5, and then try
the demo from the command line, will it load from the local ‘vexflow’ which
is vexflow 5?

Try it out and let me know what happens. I think I wrote those demos to use
the local copy of vexflow.

PR #246, opened 2024-09-21
Allow to get width without stave

### LINKED ISSUE #245: KeySignature allow to get width without assigned Stave
VexMl is crashing when trying to get the width of the element.

rvilarl:
fixes #245

see https://github.com/stringsync/vexml/pull/233

### DIFF
--- a/src/keysignature.ts
+++ b/src/keysignature.ts
@@ -129,7 +129,7 @@ export class KeySignature extends StaveModifier {
     this.formatted = false;
     return super.setStave(stave);
   }
-  
+
   // Get the `BoundingBox`
   getBoundingBox(): BoundingBox {
     if (!this.formatted) this.format();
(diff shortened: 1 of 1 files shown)


rvilarl:
@ronyeh thanks! Now setStave is used with a dummy Stave and commented. I prefer the comment because it is legitimate to use format without Stave to figure out the width.

ronyeh:
Is this code already run through eslint? I see some weird lack of spacing.
Between params and also before the brace.

Also I would just save the dummy stave to a local variable so that I don’t
need to call getStave().

rvilarl:
You are right I normally eslint. I will change it this evening.

src/keysignature.ts
```diff
@@ -240,9 +246,16 @@ export class KeySignature extends StaveModifier {
     }
   }
 
-  // Format and position the modifier
+  /**
+   *  Format and position the modifier.
```
ronyeh, line 250: Sorry I missed this in my previous review. There is an extra space between \* and Format.

ronyeh (approved):

src/keysignature.ts
rvilarl: done

PR #248, opened 2024-11-22
changed stave line drawing to place them more correctly so that the stroke-width doesn't get anti aliased

ceymard:
In the current main version, stave lines are drawn on exact integers. With a stroke-width of 1px, this means that the stroked version ends up half a pixel above and under the staff line, which ends up anti-aliased and blurry.

blurry

What is interesting is that the barlines are drawn with the width / 2 to account for it and thus also start at `y` positions that have `0.5` offsets.

This PR aims to address that : if the width is an odd number, then it will draw the staff line on a `0.5` offset, thus having its stroke nicely go from one rounded number to another.

This is the result :

nicer

The line is much crispier that way. A workaround I played with otherwise was to set `shape-rendering: crispedges` in css on `g.vf-stave path`, but while it was indeed crispier, it ended up messing with barlines vertical alignment.

Also, for similar reasons, since a stroke-width of 1.4 ends up being antialiased by the renderer I took the liberty of upping the `defaultLedgerLineStyle` to have a width of 2 since when it gets antialiased this is the number of pixels it uses anyways.

Caveat : if `scale()` is used in the renderer, we're back at having anti-aliased stuff and it's ugly again. I guess that scaling the context or the viewbox is easy, but if we want stuff to stay crisp and pixel-aligned we'd have to rewrite a buch of stuff to actually draw on pixel boundaries taking scale into account (mostly whenever `rect` or `moveTo` / `lineTo` is called)

Let me know what you think.

### DIFF
--- a/src/stave.ts
+++ b/src/stave.ts
@@ -110,7 +110,7 @@ export class Stave extends Element {
       ...options,
     };
     this.bounds = { x: this.x, y: this.y, w: this.width, h: 0 };
-    this.defaultLedgerLineStyle = { strokeStyle: '#444', lineWidth: 1.4 };
+    this.defaultLedgerLineStyle = { strokeStyle: '#444', lineWidth: 2 };
 
     this.resetLines();
 
@@ -181,11 +181,11 @@ export class Stave extends Element {
   }
 
   getTopLineTopY(): number {
-    return this.getYForLine(0) - (this.getStyle().lineWidth ?? 1) / 2;
+    return this.getYForLine(0);
   }
 
   getBottomLineBottomY(): number {
-    return this.getYForLine(this.getNumLines() - 1) + (this.getStyle().lineWidth ?? 1) / 2;
+    return this.getYForLine(this.getNumLines() - 1) + (this.getStyle().lineWidth ?? 1);
   }
 
   setX(x: number): this {
@@ -688,14 +688,17 @@ export class Stave extends Element {
     const x = this.x;
     let y;
 
+    const lineWidth = this.getStyle().lineWidth ?? 1;
+    const lineWidthCorrection = lineWidth % 2 === 0 ? 0 : 0.5;
+
     // Render lines
     for (let line = 0; line < numLines; line++) {
       y = this.getYForLine(line);
 
       if (this.options.lineConfig[line].visible) {
         ctx.beginPath();
-        ctx.moveTo(x, y);
-        ctx.lineTo(x + width, y);
+        ctx.moveTo(x, y + lineWidthCorrection);
+        ctx.lineTo(x + width, y + lineWidthCorrection);
         ctx.stroke();
       }
     }


ronyeh:
Thanks so much for investigating this! I do think that it makes sense to solve the common case (i.e., the default scenario which you fixed) and then we can figure out the general case if people need to always have it pixel aligned. Maybe someday there can be a flag to make sure staff lines and ledger lines and other vertical and horizontal lines are always pixel aligned (e.g., stem lines and bar lines).

@rvilarl and @mscuthbert do you have any feedback on this PR?

ceymard:
I think there's a lot to be investigated, really ; for instance, how do high-dpi screens fare ? I'm thinking they never even saw any problem. And the whole aligning on pixels thing is probably opening a can of worm in all the placement routines.

Fixing it for all scenari means rethinking the way almost everything is drawn and have a renderer that is pixel-aware.

Another thing that could help would be to stop stroking altogether and use instead only rects (of width 1) to draw lines. This is an approach that I tried initially but I got stuck with the `metrics.ts` file as I didn't really know which style was called for the lines (I'm thinking none and that they just use the stroke defined in the Stave context) and I wanted to keep the edits light.

I would potentially be interested in investigating that further.

mscuthbert:
Hi -- just chiming in since my comments were requested.

First -- thanks for taking Ledger line width from 1.4 to 2.0 -- I was the one who first bumped it from 1.0 to 1.4 -- I wanted to go all the way up to 2.0 (industry standard) but there was some resistance at the time.  Those ledgers look better.  No matter what happens to the rest of the PR, let's keep that.

The lines definitely look crisper and better here than before.  My Q is: have we tested this in resized canvases (both with the SVG scaling mechanisms and with CSS scaling) -- I think it should be okay, but I remember the old days where I had to add gaussian blurs to all musical examples that might scale on the web because otherwise lines would disappear when scaling. In this case, starting with a blurry line can be a help since there are at least 2 pixels to give a suggestion to a browser rendering engine that this line is important.  Blurry lines are always better than no lines.  (But crisp are best of all).

Btw -- Chrome and Safari use completely different scaling mechanisms on PNGs (JPGs/Gifs too?) which results in quite different results, so I'd want to see that it's usable on both.  Safari (Left below) blurs everything to make sure that no information is lost before scaling.  Chrome (right) looks much better and is definitely trying to keep as many thin lines thin, but this might lead to lines disappearing if the algorithm is wrong.  So let's see what happens on SVG (and Canvas if we still care about that...)  -- oh, and of course the image i'm uploading will be scaled by Github and your browser, so click to see the original if you care about actual results.

screenshot_2024-11-06_at_14 59 47

ceymard:
Ok, here are a couple of screenshot post-modification. I took them from the test (sooo useful !)

Canvas, 80% scaling, Chrome :

image

SVG, 80% scaling, Chrome :

image

Canvas, 150% scaling, Chrome :

image

SVG, 150% scaling, Chrome :

image

I have tested in webkit locally (am on Linux with the Epiphany browser) and I get similar results. What I am curious about are hi-dpi screens (I'm thinking anti-aliasing would be much nicer.)

As I was mentioning earlier : while the lines are crisp at 100% (which is I guess what we should want,) anti-aliasing and blur is back when scaling. Which could be avoided if scaling was not done by scaling the viewport or the draw surface but by taking it into account directly when creating the score.

ronyeh:
> Which could be avoided if scaling was not done by scaling the viewport or the draw surface but by taking it into account directly when creating the score.

Wow, interesting idea for sure. So maybe (someday) we'd have a pixel_perfect / pixel_aligned flag to tell VexFlow to make everything crisp, and then it would make sure a "scaled" staff is simply drawn with thicker lines or rects that always start and end on integer coordinates...

What's the display resolution and DPI on your machine? Just curious.

ceymard:
> Wow, interesting idea for sure. So maybe (someday) we'd have a pixel_perfect / pixel_aligned flag to tell VexFlow to make everything crisp, and then it would make sure a "scaled" staff is simply drawn with thicker lines or rects that always start and end on integer coordinates...

An idea I had along these lines would be to scale according to the number of pixels between two staff lines instead of giving a percentage. 

> What's the display resolution and DPI on your machine? Just curious.

I have a pretty standard 96 dpi monitor. I wish I had a retina or similar :)

mscuthbert:
Hi @ceymard - thanks for the scaling tests.  They look great!  I really don't want to put too much more work on you, but I'm afraid that 150% and 80% scaling -- being simple ratios (3:4 and 4:5) -- don't end up testing the types of bugs that come up in a "image resized to fit surrounding container" space, which would get random non-integer scaling like 73.8201% or 151.92%  (Though it is true that 50% resize and 25% resize are also VERY good test numbers to see if 1 and 2pixel lines just vanish completely).  Right now I haven't had any problem with Vexflow at those scales (while pngs are terrible at those %s)

rvilarl (approved):

rvilarl:
@ceymard thanks for the contribution.

PR #251, opened 2024-12-31
adding svg grouping to Tuplet draw method

### LINKED ISSUE #250: Grouping SVG elements for Tuplets in draw() method
Hi all, 

I noticed that the SVG elements rendered for tuplets are not grouped, unlike the components of other elements like StaveNote. Is there a specific reason for this difference? I have attached a photo of how dom renders a tuplet at the moment.

Screenshot 2024-12-21 at 3 34 19 PM

It seems that adding the following to the draw() method in tuplet.ts would group the elements effectively:

`ctx.openGroup("tuplet", this.getAttribute("id"))  // ... existing code
ctx.closeGroup()` 

This change would provide a more logical grouping of each tuplet's components and make style manipulations and such easier.

Would this be a welcome addition? If so, I would be happy to implement it and open a pull request.

Prospero23:
Fixes #250 grouping suggestion and adds simple pointer rectangle to tuplet as in StemmableNote. If adding the pointer rectangle would cause problems, its an easy delete! 

Also, moved this.setRendered() to the bottom of the class to match where it is placed in StemmableNote.

### DIFF
--- a/src/tuplet.ts
+++ b/src/tuplet.ts
@@ -296,7 +296,6 @@ export class Tuplet extends Element {
     const ctx = this.checkContext();
     let xPos = 0;
     let yPos = 0;
-    this.setRendered();
 
     // determine x value of left bound of tuplet
     const firstNote = this.notes[0] as StemmableNote;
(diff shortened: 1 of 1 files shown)


rvilarl:
Hi @Prospero23 thanks for the PR and thanks for highlighting the `pointerRect` usage in `StaveNote` @mscuthbert @AaronDavidNewman @ronyeh is that a bug? If not, why only the StaveNotes?

ronyeh:
Thanks @Prospero23!

rvilarl:
@ronyeh good to see you back! I will come back to collaboration probably next week once I am back home. Do you think that this pointRect should only be in StaveNote and Tuplet?

PR #256, opened 2025-01-27
Add complex key signatures support

rodrigos1233:
## Support for Complex Key Signatures and Alternative Definition Method

This update introduces support for complex key signatures, including theoretical or rare keys from different modes. It enhances versatility and provides additional options for defining key signatures. This implementation preserves backward compatibility, ensuring that all existing key signature definitions and functionality remain unaffected.

The feature was initially developed using the Vexflow 4 repository but has been updated and tested to ensure compatibility with Vexflow 5. I made sure the implementation works seamlessly with the updated codebase and does not introduce any breaking changes.

### Key Features

1. **Alternative Key Signature Definition:**  
   The `addKeySignature` method now accepts a new format, allowing direct specification of the number of flats or sharps:
   ```javascript
   stave.addKeySignature('flats_4');
   ```
   This adds four flats to the key signature, suitable for modes such as Ab major, F minor, Bb Dorian, C Phrygian, Db Lydian, Eb Mixolydian, or G Locrian.

2. **Support for Double Alterations:**  
   Numbers between 8 and 14 introduce double alterations (e.g., double flats or double sharps) in the key signature, enabling the rendering of more complex and less common keys.

---

This change enhances the flexibility of the `KeySignature` class, making it suitable for a wider range of musical contexts and theoretical applications.


ronyeh:
Thanks for investigating this. We are holding off on new (non-critical) PRs until V5 is shipped.

In the meantime, can you clean up the history of this PR? Right now I see 48 commits even though 3 files are changed. It also looks like there is a change to a variable name accidental vs acc.

Perhaps you forked an older (v4) version of the repo? Maybe you can branch off from the current head revision of V5 main branch?

We'll revisit to discuss this PR once it's cleaned up and after V5 is shipped.

Thanks again!

rodrigos1233:
Hi, thanks for the comment. 
Yes that is correct, I have forked from V4 and then merged the main V5 branch. I will open a new one with only my changes

rodrigos1233:
[New PR opened](https://github.com/vexflow/vexflow/pull/263)

rvilarl:
@rodrigos1233  can this PR? #263 seems to be the right one.

rvilarl:
see #263

PR #258, opened 2025-02-06
Improving documentation for demos, and fixing bugs in the demos.

ronyeh:
Improve the documentation and the demos in the main repo.

After this is merged, we can start moving most of the examples over to: 
https://github.com/vexflow/vexflow-examples

We will only keep a small set of the most import "Getting Started" demos in the main repo, to help new developers verify that their custom build works correctly.


ronyeh:
@rvilarl I will go ahead and merge this, since all the changed files are under the vexflow/demos/ directory and don't affect the core library.

The demos are now fixed and updated to work with v5.0.0.

Issue #257, opened 2025-02-05
Clean up deprecated npm packages if possible


ronyeh:
On a clean npm install, we currently get these warnings.

```
npm WARN deprecated osenv@0.1.5: This package is no longer supported.
npm WARN deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm WARN deprecated npmlog@5.0.1: This package is no longer supported.
npm WARN deprecated rimraf@2.7.1: Rimraf versions prior to v4 are no longer supported
npm WARN deprecated rimraf@2.7.1: Rimraf versions prior to v4 are no longer supported
npm WARN deprecated @humanwhocodes/config-array@0.11.14: Use @eslint/config-array instead
npm WARN deprecated rimraf@3.0.2: Rimraf versions prior to v4 are no longer supported
npm WARN deprecated are-we-there-yet@2.0.0: This package is no longer supported.
npm WARN deprecated @humanwhocodes/object-schema@2.0.3: Use @eslint/object-schema instead
npm WARN deprecated glob@7.1.7: Glob versions prior to v9 are no longer supported
npm WARN deprecated gauge@3.0.2: This package is no longer supported.
```

Need to investigate if any of these can be updated easily / safely.

ronyeh:
These seem to be sub-dependencies (e.g., **inflight** is used by **glob** which is used by **grunt**). As long as they are dev dependencies, it might be OK to leave it as is, since our build system relies on `grunt`.

rvilarl:
#259 results in:

```
npm warn deprecated osenv@0.1.5: This package is no longer supported.
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated rimraf@2.7.1: Rimraf versions prior to v4 are no longer supported
npm warn deprecated rimraf@2.7.1: Rimraf versions prior to v4 are no longer supported
npm warn deprecated glob@7.1.7: Glob versions prior to v9 are no longer supported
```

rvilarl:
We cannot do much about grunt and it does not affect our library

PR #259, opened 2025-02-07
npm-update

rvilarl:
Update of npm to reduce the number of warnings mentioned in #257


rvilarl:
I think that this merge requires no review. The update is automatic.

PR #260, opened 2025-02-08
added optional "helper" note after tuplet ratio

Prospero23:
Hey all, 
I know noncritical PRs are on hold for the moment, but still thought I would get started contributing this. Useful for notation with complex tuplets and Ferneyhough et al love this kind of thing. I have basic centering at the moment but it would be easy enough to change the layout so that the tuplet ratio itself stays centered. Also, did some small quality life improvements by creating constants for some magic numbers that were in the code. 

Very interested to see if anyone has better ideas for how to structure the API. Could it be possible to make this a flag (something like isBeatShown or showBaseDuration) and have the tuplet itself calculate what the baseNoteLength should be? Also, if it stays as an input value, I am assuming there could maybe be a better name. 

I have a question as well:
I am wanting to refactor the logic in the draw and resolveGlyphs functions into a couple of private helpers but I noticed this is done very little (if at all) in the codebase. Would it be acceptable to make public drawBracket, drawNote etc functions like what seems to be in stavenote or would it be better to just keep the functions as they are?

Thanks so much,
Ben

Screenshot 2025-02-08 at 3 44 03 PM

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -271,6 +271,10 @@ export const MetricsDefaults: Record<string, any> = {
     textYOffset: 2,
   },
 
+  TupletNote: {
+    fontSize: 12,
+  },
+
   Volta: {
     fontSize: 9,
     fontWeight: 'bold',
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -272,7 +272,7 @@ export const MetricsDefaults: Record<string, any> = {
   },
 
   TupletNote: {
-    fontSize: 12,
+    fontSize: 16,
   },
 
   Volta: {
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -269,10 +269,12 @@ export const MetricsDefaults: Record<string, any> = {
   Tuplet: {
     yOffset: 0,
     textYOffset: 2,
-  },
-
-  TupletNote: {
-    fontSize: 16,
+    bracketPadding: 5,
+    suffix: {
+      fontScale: 1 / 2,
+      extraSpacing: 1,
+      suffixOffsetY: -3,
+    },
   },
 
   Volta: {
--- a/src/tuplet.ts
+++ b/src/tuplet.ts
@@ -41,6 +41,9 @@
  *   yOffset: int, default 0
  *     manually offset a tuplet, for instance to avoid collisions
  *     with articulations, etc...
+ *
+ *   baseNoteLength: string
+ *      Show helper note duration in parenthesis after tuplet ratio.
  * }
  */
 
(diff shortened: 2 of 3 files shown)


Prospero23:
Screenshot 2025-02-08 at 7 16 29 PM

Not sure why I added the parenthesis. I think without works better and is inline with what some notation softwares like doing.

rvilarl:
Thanks a lot for the contribution. I hope we release 5 soon

Prospero23:
at the moment, i have the following in the tuplet class to resolve the new note glyphs: 
`const NOTE_GLYPHS: Record<string, string> = {
  '4': Glyphs.metNoteQuarterUp, // quarter note
  '8': Glyphs.metNote8thUp, // eighth note
  '16': Glyphs.metNote16thUp, // sixteenth note
  '32': Glyphs.metNote32ndUp, // thirty-second note
  '64': Glyphs.metNote64thUp, // sixty-fourth note
  '128': Glyphs.metNote128thUp, // one hundred twenty-eighth note
  '256': Glyphs.metNote256thUp,
  '512': Glyphs.metNote512thUp,
  '1024': Glyphs.metNote1024thUp,
};` 
I am assuming there is a much better place to put this, so I can move it wherever is best. Only kept it in the class to not step on any toes and make rebasing and such easier.

rvilarl:
Hi Ben, once again, thanks a lot for the contribution. I was not even aware that this notation existed and I found it. I have some time to make an inital review today. Do not take my comments as a must, I am also trying to figure out how to do it best. Ok I go for it.

src/tuplet.ts
```diff
@@ -41,6 +41,9 @@
  *   yOffset: int, default 0
  *     manually offset a tuplet, for instance to avoid collisions
  *     with articulations, etc...
+ *
+ *   baseNoteLength: string
```
rvilarl, line 45: Perhaps it would be better to define an unrestricted extra text, let as say `extra: string` or any other name. Doing that we let the user select which glyph to use or even if he wants to put extra characters like parenthesis.

example modifying simple text
```js
f.Tuplet({ notes: notes.slice(0, 3), options: { extra: Glyphs.metNote8thUp } });
f.Tuplet({ notes: notes.slice(3, 6), options: { extra: Glyphs.note8thUp } });
```

src/metrics.ts
```diff
@@ -271,6 +271,10 @@ export const MetricsDefaults: Record<string, any> = {
     textYOffset: 2,
   },
 
+  TupletNote: {
+    fontSize: 16,
+  },
+
```
rvilarl, line 277: Please move this into Tuplet like

```js
Tuplet: {
    yOffset: 0,
    textYOffset: 2,
    note: {
      fontScale: 1/2,
    },
  },
```
I prefer to use `fontScale` rather than `fontSize` just in case someone changes the main `fontSize`.

src/tuplet.ts
```diff
@@ -112,8 +135,10 @@ export class Tuplet extends Element {
       ratioed,
       yOffset,
       textYOffset,
+      baseNoteLength,
     };
     this.textElement = new Element('Tuplet');
+    this.noteElement = new Element('TupletNote');
```
rvilarl, line 141: If you modify the metrics you will have to call

```js
this.noteElement = new Element('Tuplet.note');
```

src/tuplet.ts
```diff
@@ -63,13 +66,30 @@ export interface TupletOptions {
...
 }
 
 export const enum TupletLocation {
   BOTTOM = -1,
   TOP = +1,
 }
 
+const BRACKET_PADDING = 5; // padding between inner text and bracket if bracket is enabled
+const NOTE_OFFSET = -3; // offset of shown note
+const EXTRA_SPACING = 1; // spacing between ratio and shown note
```
rvilarl, line 79: Metrics is the right place to define this. :)

src/tuplet.ts
Prospero23: This makes total sense to do

rvilarl (changes requested):
Comments already provided.

src/tuplet.ts
```diff
@@ -63,13 +67,18 @@ export interface TupletOptions {
...
 
 export const enum TupletLocation {
   BOTTOM = -1,
   TOP = +1,
 }
 
+// const BRACKET_PADDING = 5; // padding between inner text and bracket if bracket is enabled
+// const NOTE_OFFSET = -3; // offset of shown note
+// const EXTRA_SPACING = 1; // spacing between ratio and shown note
+
```
rvilarl, line 81: No longer required

src/tuplet.ts
```diff
@@ -104,6 +114,8 @@ export class Tuplet extends Element {
     const location = options.location || Tuplet.LOCATION_TOP;
     const yOffset = options.yOffset || Metrics.get('Tuplet.yOffset');
     const textYOffset = options.textYOffset || Metrics.get('Tuplet.textYOffset');
+    const suffix = options.suffix !== undefined ? options.suffix : '';
+
```
rvilarl, line 118: I suggest to make consistent wit the lines above
```js
const suffix = options.suffix || '';
```

src/tuplet.ts
Prospero23: whoops. fixed.

src/tuplet.ts
Prospero23: done

rvilarl (approved):

PR #264, opened 2025-02-27
Added longa, maxima, 256th, 512th and 1024th notes support

TuxIceland:
Updated Vexflow to support **longa**, **maxima**, **256th**, **512th** and **1024th** notes.
List of changes:
• Added respective glyphs for each note (also 128th); 
• Ensured each note has a correct tick number;
• Supported rests (updated rests "basic" test to include all new notes);
• Added l (Longa) and m (Maxima) aliases.
• Updated supported beam durations with 128, 256, 512, 1024.
• Added "mixed3" test on beam_tests.ts which simply renders two beams (one beam per voice) with new notes. I noticed that the last notes in the measure do not get beamed, but it doesn't seem to be caused by the added durations.

### New Rests Example
Screenshot 2025-02-27 alle 20 15 57

### Beam Example with new Durations
Screenshot 2025-02-27 alle 20 18 32

As you can see, there is a problem with the last beam notes connection. I checked if I missed something in the updated code, but I couldn't find any problem. Please tell me if I did something wrong on my side, I'm willing to help to solve this issue.

As for the rest, everything seems to work fine. Thanks.

### DIFF
--- a/src/stavenote.ts
+++ b/src/stavenote.ts
@@ -787,7 +787,6 @@ export class StaveNote extends StemmableNote {
     // addtional y shifts for rests
     let restShift = 0;
     switch (this._noteHeads[index].getText()) {
-      case Glyphs.restDoubleWhole:
       case Glyphs.restWhole:
         restShift += 0.5;
         break;
@@ -802,7 +801,14 @@ export class StaveNote extends StemmableNote {
         restShift -= 1.5;
         break;
       case Glyphs.rest128th:
+      case Glyphs.rest256th:
         restShift -= 2.5;
+        break;
+      case Glyphs.rest512th:
+        restShift -= 3.5;
+        break;
+      case Glyphs.rest1024th:
+        restShift -= 4.5;
         break;
     }
 
--- a/tests/beam_tests.ts
+++ b/tests/beam_tests.ts
@@ -33,6 +33,7 @@ const BeamTests = {
     run('Auto-stemmed Beam', autoStem);
     run('Mixed Beam 1', mixed);
     run('Mixed Beam 2', mixed2);
+    run('Mixed Beam 3', mixed3);
     run('Dotted Beam', dotted);
     run('Partial Beam', partial);
     run('Close Trade-offs Beam', tradeoffs);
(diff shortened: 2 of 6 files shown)


tests/beam_tests.ts
```diff
@@ -318,6 +319,41 @@ function mixed2(options: TestOptions): void {
...
+      'f4/32, d4/16, c4/32, c4/64, d4/128, e4/8, f4/16, e4/256, e4/512, e4/1024, e4/1024, d4/32, c4/64, c4/32, d4/16',
+      {
+        stem: 'down',
+      }
+    ),
+    { time: '31/64' }
+  );
+
+  f.Beam({ notes: voice.getTickables().slice(0, 12) as StemmableNote[] });
+  f.Beam({ notes: voice2.getTickables().slice(0, 12) as StemmableNote[] });
```
rvilarl, line 348: The slice(0,12) is limiting the beam to the first 12 notes. Remove the call to slice to include all of them.

rvilarl (changes requested):

rvilarl:
Thanks a lot for the contribution.

src/stavenote.ts
```diff
@@ -787,7 +787,6 @@ export class StaveNote extends StemmableNote {
     // addtional y shifts for rests
     let restShift = 0;
     switch (this._noteHeads[index].getText()) {
-      case Glyphs.restDoubleWhole:
```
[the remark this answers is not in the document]
TuxIceland: I noticed that in the case of a **restDoubleWhole** a shift of 0.5 would show the dot too low compared to the rest. With this configuration (restShift=0) the result is the one in the image.

tests/beam_tests.ts
TuxIceland: Stupid mistake on my hand, sorry. I will be updating this as soon as I can.

src/stavenote.ts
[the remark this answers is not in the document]
TuxIceland: Here's a before/after comparison:

### restShift = 0.5
shift0 5
 
### restShift = 0
shift0

tests/beam_tests.ts
TuxIceland: I updated **beam_tests.ts**. Everything seems to work as expected:

beam

I also updated the stave "y" from 20 to 25 so that the beam is rendered correctly.

rvilarl (approved):

Issue #265, opened 2025-03-01
For Wiki: Interactive Notes in Canvas JSFiddle


Prospero23:
Hey all, 
I made this jsfiddle as a potential helper for the wiki/documentation/vexflow-examples. Not sure where it would fit best, but I think it would be useful for new users.

It would be easy enough to add listeners for other event types, but even with just click events, the hit-detection logic is clear for people to extend as needed. 

https://jsfiddle.net/prospero23/gatjr9ky/

Also, with 4 i used the methods talked about in https://github.com/0xfe/vexflow/issues/371 for selection and such. I assume that since bounding boxes of SVGs are larger now, using canvas would probably be the better default route to go for interactivity. Is there still an easy way to use SVGs or is it better to just go with canvas?

Prospero23:
After messing around more with this, selection logic also works with clefs through getting a clefs bounding box from the stave doing something like the following (stave being a var holding the output of system.addStave() from the fiddle)
`stave.getModifiers()[2].getBoundingBox()`

I am running into issues with this working for time sig though. when i grab the time sig from the stave:
`stave.getModifiers()[3].getBoundingBox()` the x and width components seem correct but the Y and hight values seem to be set to zero.

Prospero23:
here is fiddle showing the time sig issue: https://jsfiddle.net/prospero23/fj8ecrsq/

rvilarl:
Hi Ben, the selection issue was addressed with SVG. There is an issue about that.

Prospero23:
Oh, nice! My bad for missing that, it works great. 

It would be nice to have the same logic used for notes implemented in clefs and time sigs as well. For clefs, adding the `ctx.pointerRect(...bounding box stuff)` seems to work perfectly for me. With time sigs, the y/height situation would need to be fixed first. 

I'll mess around more with it and see what I can come up with

Prospero23:
Seems that overriding getBoundingBox in timesignature does the trick. Result look good.

 I can hold off on submitting another PR until 5 fully comes out if thats best. 

<img width="465" alt="Image" src="https://github.com/user-attachments/assets/63225547-5aa6-4b8c-b702-031ab29648cf" />

rvilarl:
5 is out, please submit the PR

rvilarl:
The place for that is vexflow-examples

PR #266, opened 2025-03-06
Added pointerRects to clef and timesignature + timesig getBoundingBox override 

Prospero23:
Congrats on the release of 5.0!

as talked about in #265 

Results:
Screenshot 2025-03-06 at 12 11 21 AM

Screenshot 2025-03-06 at 12 17 35 AM

let me know what should be changed :)

I am unsure if aggregate/interchangeable time signatures need a special case. Each component is rendered as its own time signature, so maybe just having a box for each part is fine. 

interchangeable time signature structure: 
Screenshot 2025-03-06 at 12 19 36 AM
Screenshot 2025-03-06 at 12 19 49 AM


Prospero23:
Unit test that i added just combines all the different types of time signatures into one test. These already exist on clef.

I see that other components use an option`{drawBoundingBox: Bool}` for the bounding box test. I could switch from this test to do similar for each of the already existing unit tests for time signature if that is better.

src/timesignature.ts
```diff
@@ -205,23 +209,51 @@ export class TimeSignature extends StaveModifier {
...
     if (this.isNumeric) {
+      // render top text
       let startX = x + this.topStartX;
-      let y = 0;
-      if (this.botText.getText().length > 0) y = stave.getYForLine(this.topLine - this.lineShift);
-      else y = (stave.getYForLine(this.topLine) + stave.getYForLine(this.bottomLine)) / 2;
-      this.topText.renderText(ctx, startX, y);
+      if (this.botText.getText().length > 0) this.topRenderY = stave.getYForLine(this.topLine - this.lineShift);
+      else this.topRenderY = (stave.getYForLine(this.topLine) + stave.getYForLine(this.bottomLine)) / 2;
+      this.topText.renderText(ctx, startX, this.topRenderY);
```
ronyeh, line 225: Can we use braces {} for if/else blocks? I know the original was also missing braces but I feel that they generally improve readability.

ronyeh (approved):
One minor suggestion. Looks good. Thanks for the test case.

rvilarl:
@ronyeh @Prospero23 I will make a PR with the basic infraestructure. Please wait.

src/clef.ts
```diff
@@ -157,6 +157,9 @@ export class Clef extends StaveModifier {
 
     this.y = stave.getYForLine(this.line);
     this.renderText(ctx, 0, 0);
+    const bb = this.getBoundingBox();
+    ctx.pointerRect(bb.getX(), bb.getY(), bb.getW(), bb.getH());
+
```
rvilarl, line 162: This is not needed if we merge #283 first.

src/timesignature.ts
```diff
@@ -205,23 +209,51 @@ export class TimeSignature extends StaveModifier {
     this.setRendered();
     ctx.openGroup('timesignature', this.getAttribute('id'));
     this.drawAt(ctx, stave, this.x);
+    const bb = this.getBoundingBox();
+    ctx.pointerRect(bb.getX(), bb.getY(), bb.getW(), bb.getH());
```
rvilarl, line 213: This is not needed if we merge #283 first

rvilarl (changes requested):

rvilarl:
@Prospero23 #283 is already merged. Please pull and remove the unnecessary code.

Prospero23:
Everything should be fixed now.

src/timesignature.ts
Prospero23: Fixed

src/clef.ts
Prospero23: removed

src/timesignature.ts
Prospero23: removed

rvilarl (approved):

rvilarl:
@Prospero23 Thanks!

PR #283, opened 2025-04-11
pointer rect 

rvilarl:
This is just a proposal for the implementation in Element. Please feedback on approach.

Basicly:
- implementation in Element
- call once closeGroup is called

I have doubts where to configure, metrics?

### DIFF
--- a/src/element.ts
+++ b/src/element.ts
@@ -113,7 +113,7 @@ export class Element {
 
   private context?: RenderContext;
   protected attrs: ElementAttributes;
-  protected pointerRect: boolean;
+  protected shouldDrawPointerRect: boolean;
 
   protected rendered: boolean;
   protected style: ElementStyle = {};
@@ -156,7 +156,7 @@ export class Element {
     this._fontInfo = Metrics.getFontInfo(this.attrs.type);
     this.style = Metrics.getStyle(this.attrs.type);
     this.fontScale = Metrics.get(`${this.attrs.type}.fontScale`);
-    this.pointerRect = Metrics.get(`${this.attrs.type}.pointerRect`);
+    this.shouldDrawPointerRect = Metrics.get(`${this.attrs.type}.pointerRect`);
 
     // If a default registry exist, then register with it right away.
     Registry.getDefaultRegistry()?.register(this);
@@ -351,7 +351,7 @@ export class Element {
 
   /** Draw pointer rect. This allows mouse interaction with the element */
   drawPointerRect() {
-    if (this.pointerRect) {
+    if (this.shouldDrawPointerRect) {
       const bb = this.getBoundingBox();
       this.context?.pointerRect(bb.getX(), bb.getY(), bb.getW(), bb.getH());  
     }
--- a/src/element.ts
+++ b/src/element.ts
@@ -113,6 +113,7 @@ export class Element {
 
   private context?: RenderContext;
   protected attrs: ElementAttributes;
+  protected pointerRect: boolean;
 
   protected rendered: boolean;
   protected style: ElementStyle = {};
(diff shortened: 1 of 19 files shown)


rvilarl:
@ronyeh @Prospero23 @jaredjj3 thoughts?

src/canvascontext.ts
```diff
@@ -109,7 +110,8 @@ export class CanvasContext extends RenderContext {
     // Containers not implemented.
   }
 
-  closeGroup(): void {
+  closeGroup(el?: Element): void {
```
jaredjj3, line 113: If I understand correctly, you chose to hook onto `closeGroup` because it's usually the last context action when drawing. However, this breaks the encapsulation of `RenderContext` being a low level drawing tool because it now knows about `Element`.

In the `Element` subclasses we care about, is it feasible to call `this.drawPointerRect` directly in their `draw` method instead?

jaredjj3:
>I have doubts where to configure, metrics?

- If users will almost always want a global configuration, that's probably the best place today.
- If you want to support different configurations on a per-instance basis, then it could live as state on `Element` that the user has to remember to toggle.

What do you think are the likely use cases for this?

src/canvascontext.ts
rvilarl: > In the `Element` subclasses we care about, is it feasible to call `this.drawPointerRect` directly in their `draw` method instead?

yes, that was my first version but I thought it was too much duplication. But yes this breaks the encapsulation.

jaredjj3: I think the duplication is fine because the actual drawing logic is centralized in `Element.drawPointerRect`. When you make this configurable, you will need to add a conditional at each callsite, but this type of "duplication" is unavoidable and will have to live somewhere.

rvilarl: @jaredjj3 thanks, done.

Prospero23: This way looks much better!

rvilarl:
I think the use case is a global configuration.

jaredjj3:
I agree with Metrics unless...

...there's an appetite for a general config solution for vexflow (separate PR?). I found it very awkward to have to know that the values inside Metrics are cached. With a general config object, you also won't have to overload Metrics with drawing behavior config.

rvilarl:
For now in Metrics. This is an element to rethink. I believe it reduces performance.

Ready to review and merge!

jaredjj3 (approved):
LGTM

src/element.ts
```diff
@@ -347,6 +349,14 @@ export class Element {
     return this.context;
   }
 
+  /** Draw pointer rect. This allows mouse interaction with the element */
+  drawPointerRect() {
+    if (this.pointerRect) {
+      const bb = this.getBoundingBox();
+      this.context?.pointerRect(bb.getX(), bb.getY(), bb.getW(), bb.getH());
```
ronyeh, line 356: On line 354, pointerRect is a boolean. On line 356, pointerRect is a method. 

But to me, it sounds like a struct containing x,y,width,height. 

I think both should have better names.

ronyeh (changes requested):

src/element.ts
ronyeh: I'm fine with keeping pointerRect as a method in render context. 

I think the boolean should be named `shouldDrawPointerRect`.  It makes the conditional statement easier to understand.

Otherwise, this PR looks fine to me.

demos/node/customcontext.js
```diff
@@ -173,7 +173,7 @@ class CustomContext extends VF.RenderContext {
     this.log('openGroup', cls, id);
   }
 
-  closeGroup() {
+  closeGroup(this) {
```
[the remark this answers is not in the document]
rvilarl: this removed.

src/element.ts
```diff
@@ -113,6 +113,7 @@ export class Element {
 
   private context?: RenderContext;
   protected attrs: ElementAttributes;
+  protected pointerRect: boolean;
```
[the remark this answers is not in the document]
rvilarl: changed

src/element.ts
rvilarl: changed

rvilarl:
Thanks for your comments and reviews. I merge now.

PR #284, opened 2025-04-11
Update documentation to use factory not vf

mscuthbert:
Per https://github.com/vexflow/vexflow/discussions/278

See parallel PRs at
* https://github.com/vexflow/vexflow-examples/pull/14 
* https://github.com/vexflow/vexflow-sandbox/pull/2

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,13 +31,12 @@ npm install vexflow
 Factory and EasyScore are VexFlow's high-level API for creating staves, voices, and notes. On a web page containing a `<div id="output"></div>`, the following code displays a score:
 
 ```javascript
-const { Factory } = VexFlow;
-const f = new Factory({
+const factory = new VexFlow.Factory({
   renderer: { elementId: 'output', width: 500, height: 200 },
 });
 
-const score = f.EasyScore();
-const system = f.System();
+const score = factory.EasyScore();
+const system = factory.System();
 
 system
   .addStave({
@@ -49,7 +48,7 @@ system
   .addClef('treble')
   .addTimeSignature('4/4');
 
-f.draw();
+factory.draw();
 ```
 
 [See it running here.](https://vexflow.github.io/vexflow-examples/getting-started)
--- a/demos/entry/vexflow.html
+++ b/demos/entry/vexflow.html
@@ -22,12 +22,12 @@
     <pre id="code"></pre>
     <div id="output"></div>
     <script>
-      const sampleCode = `const vf = new VexFlow.Factory({
+      const sampleCode = `const factory = new VexFlow.Factory({
   renderer: { elementId: 'output', width: 300, height: 180 },
 });
 
-const score = vf.EasyScore();
-const system = vf.System();
+const score = factory.EasyScore();
+const system = factory.System();
 
 system
   .addStave({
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -31,13 +31,12 @@ npm install vexflow
 Factory and EasyScore are VexFlow's high-level API for creating staves, voices, and notes. On a web page containing a `<div id="output"></div>`, the following code displays a score:
 
 ```javascript
-const { Factory } = VexFlow;
-const f = new Factory({
+const factory = new VexFlow.Factory({
   renderer: { elementId: 'output', width: 500, height: 200 },
 });
```
ronyeh, line 36: Do you think this is better?
```
const Factory = VexFlow.Factory;
const factory = new Factory(...);
```

I'm thinking back to my newbie programmer years and I think I would be confused by the dot in `new VexFlow.Factory()`. I always remember learning `new ClassName()` and have never seen `new NameSpace.ClassName()`. 

Although my version is also slightly confusing because there is a capitalized Factory (Class) and a lowercase factory (local var). 

Novice programmers can be confused by many things!

demos/entry/vexflow.html
```diff
@@ -39,7 +39,7 @@
   .addClef('treble')
   .addTimeSignature('4/4');
 
-vf.draw();`;
+factory.draw();`;
```
ronyeh, line 42: In general I approve of writing out `factory` instead of f or vf or vff.

However I'm not a huge fan of seeing `new VexFlow.ClassName()` everywhere. It can get wordy especially if we use multiple VexFlow classes. 

I think that's why we "extracted" the class names at the top of each example. The destructure assignment acts like an import at the top, so we don't have to have VexFlow.* repeated throughout the code.

mscuthbert: I don't think I ever added `Vexflow.` anywhere that there was more than one classname extracted.  I specifically left alone `const { Dot, Factory } = VexFlow` and `const { Factory, StaveHairpin } = VexFlow`.   Unless I made a mistake, I thought I only used `new Vexflow.Factory()` when Factory was the only attribute of VexFlow that was referenced.

README.md
mscuthbert: I don't think that the `new VexFlow.Factory()` is likely to be a main source of confusion (as you noted, Factory vs. factory is harder, and at least one is namespaced).

I think more likely to be of confusion is the double nested object in the call to Factory -- why is there `({` and not just `(` etc. (I know, and you know, but we're used to immediate abstractions and having call stacks in our heads).

If the dot in `VexFlow.Factory()` is likely to be confusing, then the learner has no chance to figure out what is happening in the dotted chain call below it with the add clefs, etc. and nested created objects.  That's the place I think confusion is more likely to happen.

(This `f` I would've left alone if I wasn't making a complete change of `vf` to `factory` btw. -- my main confusion always was thinking that `vf` referred to the VexFlow library or some sort of `new VexFlow()` object.

ronyeh: Sounds good to me. 

I'm fine with `new VexFlow.Factory(...)` for beginner focused tutorials. 

But if we have advanced tutorials importing lots of classes, we should strive for conciseness to avoid littering the examples with VexFlow.* everywhere.

ronyeh (approved):

PR #293, opened 2025-04-30
added noImplicitOverride to tsconfig

Prospero23:
Hey all,

I noticed that noImplicitOverride (https://www.typescriptlang.org/tsconfig/#noImplicitOverride) was commented out in the project's tsconfig.json with a note about wanting it to be added in the future. Well, maybe the future is now! 

The non tsconfig modified files are just adding `override` wherever the ts compiler was telling me there was an error with the new rule.

### DIFF
--- a/src/accidental.ts
+++ b/src/accidental.ts
@@ -53,7 +53,7 @@ export class Accidental extends Modifier {
   protected cautionary: boolean;
 
   /** Accidentals category string. */
-  static get CATEGORY(): string {
+  static override get CATEGORY(): string {
     return Category.Accidental;
   }
 
@@ -521,7 +521,7 @@ export class Accidental extends Modifier {
     this.reset();
   }
 
-  protected reset(): void {
+  protected override reset(): void {
     this.text = '';
 
     if (!this.cautionary) {
@@ -540,7 +540,7 @@ export class Accidental extends Modifier {
   }
 
   /** Attach this accidental to `note`, which must be a `StaveNote`. */
-  setNote(note: Note): this {
+  override setNote(note: Note): this {
     defined(note, 'ArgumentError', `Bad note value: ${note}`);
 
     this.note = note;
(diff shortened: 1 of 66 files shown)


rvilarl:
Based on what I read, see below, I would merge the proposal. @ronyeh @jaredjj3 what do you think?

_"
Set noImplicitOverride: true
It lets Typescript do more of what typescript is meant to do.
And its usually a quick and easy increase of type-safety in your codebase.

[#noImplicitOverride](https://www.typescriptlang.org/tsconfig#noImplicitOverride)

Here are 3 reasons why you should set tsconfig.json #noImplicitOverride

Prevents Overrides that don't actually override:
It prevents your developers from mistakenly thinking they are overriding something when they are not, preventing errors like spelling mistakes or attempting to override a removed/updated method/variable on the base class.

Ensures Up-to-Date Sub-classes:
Lets Typescript give an error if a subclass becomes outdated due to changes in the base class, such as the removal or modification of a method.

Requires Explicit Method Signature:
The override keyword shows clearly what the intention
of the method is. If it's not there you would have to go to the base class (and potentially to its base and so forth) to see if the method is overridden or not
"_

jaredjj3:
Sure, a drawback is that you'll have to add `override` everywhere. Otherwise, SGTM.

sschmidTU:
*Just my mostly useless opinion as this is already merged and I'm not voting for reverting (as that also has costs):*
This looks very ugly to me, and an unnecessary huge set of changes to the code (which makes comparing commits/branches difficult).
Why do we need override for a draw method that returns void?
image

But it's okay, not a big deal, just a lot of boilerplate added.

ronyeh:
I understand the concerns with adding visual clutter to the code base with `override` keywords. But in the end I agree that the benefits of requiring the keyword outweigh the costs.

1) If someone accidentally writes a method that shadows a parent class's method, this setting will throw an error. 

2) In our current new world of LLMs, having the keyword will give extra hints to the AI agent. Also, the developer may not need to type `override` since the AI autocomplete will print out the code for us. ;-)

3) The compiled library is just JS, so if someone is writing in pure JS, this requirement has no negative impact. I sometimes write in JS if I'm throwing together a quick prototype with a `<script>` tag inside a single `index.html`. In that case, I would just include the CDN version of VexFlow, and I wouldn't have to use the override keyword when subclassing and overriding methods.

PR #300, opened 2025-06-15
clean and add bounding box override to tuplet

Prospero23:
To allow `getBoundingBox()` to work, I have made the text and suffix elements have the correct bounds information be published during `draw()`. I am treating the tuplet's `this.x`, `this.y`, `this.height`, and `this.width` as the bounds of a potentially drawn bracket. Not sure if this is the best solution... but it allows the three elements (bracket, suffix, and text) to all cleanly store their bounds information to be merged in `getBoundingBox()`.

 I believe the tests I added cover all tuplet cases. Happy to answer any questions. 

I also continued to slightly rid the class of magic numbers by moving bracket leg length and bracket line width to metrics. 

Screenshot 2025-06-15 at 12 13 04 PM

### DIFF
--- a/src/metrics.ts
+++ b/src/metrics.ts
@@ -275,7 +275,11 @@ export const MetricsDefaults: Record<string, any> = {
     pointerRect: true,
     yOffset: 0,
     textYOffset: 2,
-    bracketPadding: 5,
+    bracket: {
+      padding: 5,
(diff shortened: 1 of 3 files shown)


ronyeh:
thanks for removing magic numbers! :-)

PR #305, opened 2025-07-28
Fix svg group: Include class attribute

remileduc:
Originally from @rednebmas, but on the old repo: https://github.com/0xfe/vexflow/pull/1656

Actually use the `class` attribute on drawable elements. Currently we can set it with `Element.addClass()` but it is totally ignored at rendering.
The idea is, for instance, to be able to colorize some notes or ease the interactivity with some `Elements`.


rvilarl:
@remileduc thanks a lot

PR #263, opened 2025-02-11
Add complex key signatures

rodrigos1233:
## Support for Complex Key Signatures and Alternative Definition Method

[Original PR ](https://github.com/vexflow/vexflow/pull/256), remade from main V5 branch following comment

This update introduces support for complex key signatures, including theoretical or rare keys from different modes. It enhances versatility and provides additional options for defining key signatures. This implementation preserves backward compatibility, ensuring that all existing key signature definitions and functionality remain unaffected.

### Key Features

1. **Alternative Key Signature Definition:**  
   The `addKeySignature` method now accepts a new format, allowing direct specification of the number of flats or sharps:
   ```javascript
   stave.addKeySignature('flats_4');
   ```
   This adds four flats to the key signature, suitable for modes such as Ab major, F minor, Bb Dorian, C Phrygian, Db Lydian, Eb Mixolydian, or G Locrian.

2. **Support for Double Alterations:**  
   Numbers between 8 and 14 introduce double alterations (e.g., double flats or double sharps) in the key signature, enabling the rendering of more complex and less common keys.

There are 2 ways to display double alterations on a key signatures that can be found on the web: putting the altered note in the original position (F C G D A E B for sharps, reverse for flats) or shifting the double alterations to the end of the key (C G D A E F for 1 double sharp). The latter seems to be the most used, for example on [wikipedia](https://en.wikipedia.org/wiki/D-flat_minor), hence why I chose this way, but it can be changed easily, on line 544 of tables.ts

### Image examples
```javascript
//Use case example: D flat minor
stave
  .addClef('treble')
  .addTimeSignature('4/4')
  .addKeySignature('flats_8');
```
Capture d’écran 2025-02-24 à 21 48 52

```javascript
//Use case example: G sharp major
stave
  .addClef('treble')
  .addTimeSignature('4/4')
  .addKeySignature('sharps_8');
```
Capture d’écran 2025-02-24 à 21 59 27

```javascript
//Use case example: D flat locrian
stave
  .addClef('treble')
  .addTimeSignature('4/4')
  .addKeySignature('flats_10');
```
Capture d’écran 2025-02-24 à 22 08 04

```javascript
stave
  .addClef('treble')
  .addTimeSignature('4/4')
  .addKeySignature('sharps_14');
```
Capture d’écran 2025-02-24 à 22 08 51

```javascript
//Example use case: E flat major
stave
  .addClef('treble')
  .addTimeSignature('4/4')
  .addKeySignature('flats_3');
```
[...]

PR #316, opened 2026-02-11
Key signatures new syntax fix

rodrigos1233:
This PR addresses a bug introduced in [PR 263](https://github.com/vexflow/vexflow/pull/263). 

While trying to use this new implementation, I noticed that a typo prevented to use key signatures with flat with the new syntax. 

This PR now corrects it. I also updated the documentation to more accurately reflect the possibilities of the implementation.


0xfe:
Looks good, thanks.

PR #321, opened 2026-05-16
Review code quality and annotate source with inline code-review markers.

ronyeh:
I had some Claude extra usage tokens, so I asked it to review the current state of VexFlow to look for bugs, typos, and bad (or outdated) comments.

I'll leave it here as a draft, but the output looks pretty good so I'll break it up into multiple PRs.

The goal is to make VexFlow bug free :-). No new features. Just make it a nice and clean library.

## More info

Pass over every reviewable file in src/ and entry/ to flag possible bugs and incorrect/outdated comments. Markers used:
- `// POSSIBLE_BUG_BELOW?` for suspect code
- `// BAD_COMMENT_BELOW?` for stale/wrong comments
- `// SUGGESTED_FIX` for proposed replacements (kept as comments; original code preserved beneath)

Original logic untouched — only review annotations added. Full table of findings with file:line references is in FINDINGS.md.

Skipped: src/glyphs.ts (5880 lines of pure glyph data) and src/fonts/*.

@ronyeh: I used Claude Opus 4.7 to conduct this review.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,21 @@
+# Code Quality and Comments
+
+When reading through code, if you notice there are incorrect or outdated comments, please mark them so that we can review the comments:
+
+```
+  // BAD_COMMENT_BELOW?
+  // renderSymbol(...) takes three arguments.
+  function renderSymbol(arg1) {
+    ...
+  }
+```
+
+When reading through code, if you notice possible bugs, please mark them as follows:
+
+```
+  // addTwoNumbers(...)
+  function addTwoNumbers(a, b) {
+    // POSSIBLE_BUG_BELOW?
+    return a - b;
+  }
+```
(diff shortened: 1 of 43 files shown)


mscuthbert:
This looks great to me -- wow -- those bugs are real (the first suggestion of how to fix is almost certainly incorrect though -- we want to do something with notes.keys[n] not remove the loop).

Maybe we want to have a broader discussion about use of Agents in improving Vexflow -- I think this is exactly the kind of code where they'd be very valuable.  And we might also want to have a period of "backwards incompatible is okay" to do things like fix curTransfrom to curTransform without committing to permanent aliases for misspelled items.

My suggestion is to not check in the ai-code-review and esp. the entire Claude log with reviews (but maybe point to it or something) -- otherwise over time 90%+ of the VF repo will be AI logs.

It would also be good to move CLAUDE.md to AGENTS.md and put a symlink from AGENTS.md to CLAUDE.md -- that seems to be the growing consensus of how to open up to other agents (all the rest read AGENTS.md; only CLAUDE as market leader is able to not follow standards).  Similarly if we end up with reusable SKILL.md files, putting them in .agents/skills and symlinking that to .claude/skills  is a good way to have them be available throughout.

Thanks @ronyeh for doing this great work and getting inertia going in the right direction again!

PR #324, opened 2026-08-05
Add source-scoped formatter layout padding

bejomin:
## Summary

- add directional, source-scoped layout padding to notes and tick contexts
- preserve hard minimum distances during formatter justification
- allow a caller to replace or remove its own padding without disturbing other constraints
- leave existing score layout unchanged when no external padding is supplied

This gives notation consumers a typed way to express collision clearances without folding hard constraints into rhythmic spacing or repeatedly accumulating padding across rerenders.

## Validation

- `npm run lint` under Node 24
- `PUPPETEER_EXECUTABLE_PATH=/usr/bin/google-chrome-stable npm test` under Node 24
- 1,862 QUnit tests passed

### DIFF
--- a/src/formatter.ts
+++ b/src/formatter.ts
@@ -503,6 +503,24 @@ export class Formatter {
     return this.minTotalWidth;
   }
 
+  /**
+   * Return the exact portion of the minimum width contributed by layout-only
+   * TickContext padding. Call after preCalculateMinTotalWidth() or format().
+   *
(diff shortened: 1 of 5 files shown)


bejomin:

- 1,862 tests passed
- 7,671 assertions passed
- The test count is 2 higher than the baseline, so this PR introduces 2 new tests; they do not add new rendered image fixtures.

bejomin:

PR #322, opened 2026-07-28
Fix positioned articulation origins

bejomin:
## Summary
This fixes positioned articulations that drift to the wrong notehead/stem origin after the VexFlow 5 changes around notehead and modifier positioning.

## Validation
- `npm run lint`
- `PUPPETEER_EXECUTABLE_PATH=/usr/bin/google-chrome-stable npm test`

PR #325, opened 2026-08-05
Expose finalized notation geometry

bejomin:
## Summary

- expose exact selected-notehead bounds before drawing
- calculate articulation geometry idempotently after final beam/stem layout
- support absolute outward articulation displacement without rerender drift
- expose the same finalized beam polygons and tie curves consumed by drawing

These APIs let notation-layout clients perform collision detection against the geometry VexFlow will actually render, rather than approximating whole-note bounding boxes or duplicating draw-time calculations.

## Dependency

This branch is stacked on #322 because stable articulation bounding boxes require the positioned-origin correction. Once #322 merges, this PR can be rebased onto `main` and will contain only the three finalized-geometry commits.

## Validation

- VexFlow lint under Node 24
- 1,880 tests passed, including repeated-layout, final-beam, selected-head, beam-polygon, and tie-curve coverage

### DIFF
--- a/src/articulation.ts
+++ b/src/articulation.ts
@@ -2,6 +2,7 @@
 // @author Larry Kuhns.
 // MIT License
 
+import { BoundingBox } from './boundingbox';
 import { Builder } from './easyscore';
 import { Glyphs } from './glyphs';
 import { Modifier } from './modifier';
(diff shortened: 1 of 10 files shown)


bejomin:

- 1,880 tests passed
- 7,742 assertions passed
- The test count is 20 higher than the baseline, so this PR introduces 20 new tests, producing 17 new visual fixture images.

bejomin:

PR #326, opened 2026-08-05
Make tuplet redraw positioning idempotent

bejomin:
## Summary

- keep the tuplet text element's stored position relative to its owning tuplet
- avoid feeding a previously rendered absolute coordinate back into the next draw
- add a black-box repeated-draw test that verifies the tuplet number remains fixed

This makes repeated rendering idempotent without changing first-render placement.

## Validation

- VexFlow lint under Node 24
- 1,864 tests passed

### DIFF
--- a/src/tuplet.ts
+++ b/src/tuplet.ts
@@ -385,18 +385,17 @@ export class Tuplet extends Element {
 
     // draw ratio text (x:y)
     let currentX = notationStartX;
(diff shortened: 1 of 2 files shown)


bejomin:

- 1,864 tests passed
- 7,666 assertions passed
- The test count is 4 higher than the baseline, so this PR introduces 4 new tests, producing 4 new visual fixture images.

bejomin:

PR #327, opened 2026-08-05
Refine grace-note slur geometry

bejomin:
## Summary

- route grace-note slurs from finalized selected-head and stem geometry
- keep articulation layout stable when grace slurs share an endpoint
- expose deterministic source-spanning grace slur geometry for collision handling
- clear grace beams with a compact, configurable visual gap
- retain draw/layout idempotence across repeated rendering

## Dependency

This branch is stacked on #325, which supplies the finalized notehead, articulation, beam, and tie geometry used here. #325 is in turn stacked on #322. The dependency commits can be removed by rebasing as those prerequisite PRs merge.

## Validation

- VexFlow lint under Node 24
- 1,880 tests passed, including stem-attached, beamed, source-spanning, and repeated-layout grace slurs

### DIFF
--- a/src/articulation.ts
+++ b/src/articulation.ts
@@ -2,6 +2,7 @@
 // @author Larry Kuhns.
 // MIT License
 
+import { BoundingBox } from './boundingbox';
 import { Builder } from './easyscore';
 import { Glyphs } from './glyphs';
 import { Modifier } from './modifier';
@@ -19,6 +20,15 @@ export interface ArticulationStruct {
   aboveCode?: string;
   belowCode?: string;
   betweenLines: boolean;
+}
+
+export interface ArticulationLayout {
+  x: number;
+  y: number;
+  boundingBox: BoundingBox;
+  position: number;
+  index: number;
+  outwardShift: number;
 }
 
 // eslint-disable-next-line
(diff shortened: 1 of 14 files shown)


bejomin:
Again, I've looked over the proposed source changes and they seem reasonable. Please correct me if my AI friend and I are breaking with any previous layout/rendering decisions.

As you may be able to guess, I've been porting OSMD over to the latest VexFlow and trying to improve the rendered output in the process.

AaronDavidNewman:
I see now, this is a superset of 325 and 322.  I see the difference in grace notes now and that looks good.

I'm not quite sure I understand the changes to articulation:

image

Does this means that SVG is behaving differently than canvas?

bejomin:
It appears that's what VexFlow does natively, the one Canvas test and the three SVG tests. Seems strange to me, but ChatGPT matched that behaviour when it put this together for me. I suppose it would be fairly quick to expand the testing to establish parity between the two?

bejomin:
Completed working through the visual regression tests and came up with a few problems, which are now fixed. Here's an image that shows the before and after of these changes.
04-canvas_Grace_Notes Grace_Note_Basic_with_Slurs Bravura

bejomin:

